PR #1, opened 2018-07-18
EMP-271 phase1 chroot into customized rootfs

raysiudak:


build.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/sh
+
+set -eu
+
+is_substring()
+{
+    case "$2" in
+        *$1*)
```
oliv3r: can you reduce the indentation by one on this line please

build.sh
```diff
@@ -0,0 +1,124 @@
...
+        return 0
+        ;;
+        *)
```
oliv3r: dito

build.sh
oliv3r: what happens if i do is_substring "" "string"?

probably not important however ;)

build.sh
```diff
@@ -0,0 +1,124 @@
...
+    esac
+}
+
+OK=0
+CURRENT_DIR=$(pwd)
+MY_NAME=$(basename -- "$0")
+ROOTFS_ARCHIVE="rootfs.tar.xz"
+BUILD_DIRECTORY="${CURRENT_DIR}/build"
+ROOTFS_DIRECTORY="${BUILD_DIRECTORY}/rootfs"
+DEBIAN_VERSION="jessie"
```
oliv3r: i think it's better to have your variables above boiler plate functions
(and boiler plate variables below these, such as 'OK'
also, can you make them DEBIAN_VERSION="${DEBIAN_VERSION:-jessie}" where applicable so that they can be overrideable

build.sh
```diff
@@ -0,0 +1,124 @@
...
+        ;;
+        *)
+        return 1
+        ;;
+    esac
+}
+
+OK=0
+CURRENT_DIR=$(pwd)
+MY_NAME=$(basename -- "$0")
```
oliv3r: what does the -- do here? Why is it needed?
also, i'm surprised that shellcheck did not complain here, but while unlikely, isn't the returned value a string? E.g. it could be "My name.sh" and thus contain spaces, so it's safer to quote the result.

Finally, would you mind renaming it to PROGRAM_NAME? my name isn't build.sh, it's @oliv3r :p

build.sh
```diff
@@ -0,0 +1,124 @@
...
+ROOTFS_ARCHIVE="rootfs.tar.xz"
+BUILD_DIRECTORY="${CURRENT_DIR}/build"
```
oliv3r: to be consistent, either use DIRECTORY or DIR. Where DIR has my personal preference, it's nice and short and a well established abbreviation :) (even BUILD_DIR and CUR_DIR would be okay with me) but importantly immediately clear what it means.

build.sh
```diff
@@ -0,0 +1,124 @@
...
+ROOTFS_DIRECTORY="${BUILD_DIRECTORY}/rootfs"
+DEBIAN_VERSION="jessie"
+
+cleanup()
+{
+    if is_substring "${CURRENT_DIR}" "${BUILD_DIRECTORY}"
+    then
```
oliv3r: up until now we used if <test>; then which saves on vertical linespace and puts the if and fi above eachother

build.sh
oliv3r: if we want to keep it as part of our CWD, should we not prefer to make it a hidden dir? further more i think it's nicer to also encode the target architecture here, .build_${ARCH}" which makes us more future proof for an imx8 for example that's aarch64 and not compatible

build.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/sh
+
+set -eu
+
+is_substring()
+{
+    case "$2" in
+        *$1*)
+        return 0
```
oliv3r: please use
```sh
return false
```
instead

build.sh
```diff
@@ -0,0 +1,124 @@
...
+        ;;
+        *)
+        return 1
```
oliv3r: dito with true

build.sh
```diff
@@ -0,0 +1,124 @@
...
+CURRENT_DIR=$(pwd)
+MY_NAME=$(basename -- "$0")
+ROOTFS_ARCHIVE="rootfs.tar.xz"
+BUILD_DIRECTORY="${CURRENT_DIR}/build"
+ROOTFS_DIRECTORY="${BUILD_DIRECTORY}/rootfs"
+DEBIAN_VERSION="jessie"
+
+cleanup()
+{
+    if is_substring "${CURRENT_DIR}" "${BUILD_DIRECTORY}"
```
oliv3r: what if by whatever reason CUR_DIR and BUILD_DIR are both '/'
is_substr will return true, and thus we will remove rm -rf / :)

more importantly, you are not after is_substr, but startswith(). The build directory should start with the CWD as its prefix.
and then the function may not even be that important anymore:
```sh
if [ "${BUILD_DIR#${CUR_DIR}}" != "${BUILD_DIR}" ]; then
    rm -rf
fi
```
The # operator strips the smallest prefix matching whatever follows it. e.g.
${FULL_STRING#STRING_TO_BESTRIPED_BEFORE}
thus /home/jedi/build#/home/jedi becomes just '/jedi' which no longer matches ${BUILD_DIR}
all other cases, there will be not a prefix match to strip and thus the strings will be identical.

This also removes the requirement for a second check, to see if both strings are not equal (the / = / problem) via
```sh
if is_substring "${CUR_DIR}" "${BUILD_DIR}" && [ "${#BUILD_DIR}" -ne "${#CUR_DIR}" ]; then
```
which i think is a little harder on the eyes/more to process mentally

See http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html search for "Remove Smallest Prefix Pattern"

~~Alternatively can we not build in a temporary directory? What is the advantage of building under our current dir?
e.g.~~
```sh
BUILD_DIR="$(mktemp -d)"
...
       if [ "$(dirname "${BUILD_DIR}")" != "/tmp" ]; then
               exit 1
       fi
       rm -rf "${BUILD_DIR}"
```
~~this ensures that we are always deleting something only from /tmp.~~

build.sh
```diff
@@ -0,0 +1,124 @@
...
+    then
+        rm -rf "${BUILD_DIRECTORY}"
+    fi
+}
+
+prepare()
```
oliv3r: while i do this plenty; but prepare_what? :)

build.sh
```diff
@@ -0,0 +1,124 @@
...
+{
+    if [ ! -d "${ROOTFS_DIRECTORY}/usr/bin" ]
+    then
+        mkdir -p "${ROOTFS_DIRECTORY}/usr/bin"
+    fi
+
+    if [ ! -f "${ROOTFS_DIRECTORY}/usr/bin/qemu-arm-static" ]
```
oliv3r: can we maybe turn QEMU_STATIC_BIN="$(command -V qemu-${ARCH}-static)" into a variable at the top?

also -x rather then -f (as we need to execute it) not sure if the user running the script (e.g. not root) needs to match the user running the command (root in the chroot) e.g. doing a test -x <file> as user, checks is that user can run it vs the different (root) user running it eventually

build.sh
```diff
@@ -0,0 +1,124 @@
...
+    then
+        cp /usr/bin/qemu-arm-static "${ROOTFS_DIRECTORY}/usr/bin/"
```
oliv3r: i'm not sure how to solve this better, as this right now requires us to copy the file into the chroot.

btw, did you try doing a mount --bind "${QEMU_STATIC_BIN}" "${ROOTFS_DIR}/usr/bin/" into the chroot instead? (it's a docker trick which i think works really well.

build.sh
```diff
@@ -0,0 +1,124 @@
...
+
+bootstrap_rootfs()
+{
+    printf "Bootstrapping rootfs to %s\\n" "${ROOTFS_DIRECTORY}"
+    debootstrap \
+        --arch=armhf \
+        --variant=minbase \
+        --include=f2fs-tools,mtd-utils,busybox,udisks2,rsync \
+        "${DEBIAN_VERSION}" "${ROOTFS_DIRECTORY}" http://ftp.debian.com/debian
+}
```
oliv3r: note to self, if it turns out a debian base is to large; we can bootstrap an alpine image instead, which should be < 10 mb. Shall we create a new issue to investigate at some point? (after completion of the entire current update stack we are working on now) I have a gut feeling that an alpine image may be more suitable to what we need looking at the current image size.

build.sh
```diff
@@ -0,0 +1,124 @@
...
+        cp /usr/bin/qemu-arm-static "${ROOTFS_DIRECTORY}/usr/bin/"
+    fi
+}
+
+
+bootstrap_rootfs()
+{
+    printf "Bootstrapping rootfs to %s\\n" "${ROOTFS_DIRECTORY}"
+    debootstrap \
+        --arch=armhf \
```
oliv3r: you are doing an armhf bootstrap without the foreign/second stages (which i think is better). However this _requires_ that we have qemu-${ARCH}-static in the path. So you should check that when starting this function as a requirement.

build.sh
```diff
@@ -0,0 +1,124 @@
...
+strip_rootfs()
+{
+    printf "Stripping rootfs\\n"
+    if [ -f "${ROOTFS_DIRECTORY}/usr/bin/qemu-arm-static" ]
+    then
+        printf "Removing qemu\\n"
+        # TODO: Strip even more see issue EMP-324 Reduce rootfs size
+        rm -f "${ROOTFS_DIRECTORY}/usr/bin/qemu-arm-static"
+    fi
+}
```
oliv3r: with a bind mount, this simplifies to only umount :) in an 'unprepare()'

build.sh
```diff
@@ -0,0 +1,124 @@
...
+    if [ ! -f "${ROOTFS_DIRECTORY}/usr/bin/qemu-arm-static" ]
+    then
+        cp /usr/bin/qemu-arm-static "${ROOTFS_DIRECTORY}/usr/bin/"
+    fi
+}
+
+
+bootstrap_rootfs()
+{
+    printf "Bootstrapping rootfs to %s\\n" "${ROOTFS_DIRECTORY}"
```
oliv3r: why switch to printf from echo [0] as that's just as portalble? (printf has its uses for sure, but overkill here?) also why is \n to be double escaped?

that said i'm a printf fan myself as its more programmy :p

[0] http://pubs.opengroup.org/onlinepubs/009695399/utilities/echo.html

build.sh
```diff
@@ -0,0 +1,124 @@
...
+        rm -f "${ROOTFS_DIRECTORY}/usr/bin/qemu-arm-static"
+    fi
+}
+
+compress_rootfs()
+{
+    if [ -f "${BUILD_DIRECTORY}/${ROOTFS_ARCHIVE}" ]
+    then
+        rm -f "${BUILD_DIRECTORY}/${ROOTFS_ARCHIVE}"
+    fi
```
oliv3r: This will not work if we want versioned files (e.g. branch + commitish + date or something). as it will only remove our current artifact.

i ran into a similar issue so maybe we can do that here as well; note that i'm not happy with the file filter there but I wanted to keep it generic and applicable to all our packages. What would have helped maybe would have been ${project}*${arch}*.deb or something. Furthermore This loop deletes the local symlink to the archive and wherever the symlink points to in the build directory.

also i don't remember why i have asterisks on the in and outside of the brackets anymore. typo?
```sh
for debpkg in *"${ARCH}*".deb; do
	if [ "${debpkg}" = "*${ARCH}*.deb" ]; then
		continue
	fi
	pkglink="$(readlink "${debpkg}")"
	if [ -L "${pkglink}" ]; then
		echo "Removing ${pkglink}"
		rm "${pkglink}"
	fi
	if [ -e "${debpkg}" ]; then
		echo "Removing ${debpkg}"
		rm "${debpkg}"
	fi
done
```

build.sh
```diff
@@ -0,0 +1,124 @@
...
+
+    printf "Compressing rootfs\\n"
+    tar -cJf "${BUILD_DIRECTORY}/${ROOTFS_ARCHIVE}" -C "${ROOTFS_DIRECTORY}" .
+    chown "$(id -un):$(id -gn)" "${BUILD_DIRECTORY}/${ROOTFS_ARCHIVE}"
```
oliv3r: are we not running the build as user? and as such, isn't the file owned by the user who ran the job?
build_for_ultimaker.sh script should account for that iirc

actually, you are requesting the id of the current user, which is the default permission already anyway, so this just repeats what already is no? unless we run the script as sudo, but only if sudo is not overriding the login shell and thus the uid if i'm not mistaken.

build.sh
```diff
@@ -0,0 +1,124 @@
...
+    printf "Created %s\\n" "${ROOTFS_ARCHIVE}"
+}
+
+usage()
+{
+cat <<-EOT
+Usage:   ./${MY_NAME} [ -h | -c ]
+-c   Cleanup, after a run you have the possibility to get the artifact and then cleanup
+-h   Print usage
+NOTE: This script should be run as root
```
oliv3r: it's okay to indent the content of the here-document 1 tab to make it more obvious what is part of the here document.

additionally, for readability it be nice to indent the options.

Also, you list your arguments as -h OR -c, are they mutually exclusive? I doubt it. Just say [OPTIONS] and optinally put it under an 'options' header in the output

build.sh
```diff
@@ -0,0 +1,124 @@
...
+    printf "Compressing rootfs\\n"
+    tar -cJf "${BUILD_DIRECTORY}/${ROOTFS_ARCHIVE}" -C "${ROOTFS_DIRECTORY}" .
+    chown "$(id -un):$(id -gn)" "${BUILD_DIRECTORY}/${ROOTFS_ARCHIVE}"
+    printf "Created %s\\n" "${ROOTFS_ARCHIVE}"
+}
+
+usage()
+{
+cat <<-EOT
+Usage:   ./${MY_NAME} [ -h | -c ]
```
oliv3r: ah so that's why you strip the path with basename. i'd recommend against that.
if I have the script in my path, i just run it and it will output it properly. I think it is actually expected that way and very common:
```
root@ultimakersystem-ccbdd3004868:~# /bin/ls --help
Usage: /bin/ls [OPTION]... [FILE]...


root@ultimakersystem-ccbdd3004868:~# ls --help
Usage: ls [OPTION]... [FILE]...

root@ultimakersystem-ccbdd3004868:~# `which ls` --help
Usage: /bin/ls [OPTION]... [FILE]...
```

also it now requires to have this script to be installed in ./ and only be called as ./ even if it is in my path, which seems a bit silly :)

~~BTW, why not just use "${0}" I don't think you can (easily) overwrite it. Though I think it's good style to put $0 in PROG_NAME at the start and refer to that never the less. So actually ignore this :p~~

build.sh
oliv3r: Note that the script requires root permissions (which hopefully may not be true anymore if we'd use alpine/something self build), which is something the users has to resolve, su root; sudo <script>; suid or via setcap. and that IS a requirement, permissions, not that it _should_ be run as root.

build.sh
```diff
@@ -0,0 +1,124 @@
...
+-c   Cleanup, after a run you have the possibility to get the artifact and then cleanup
+-h   Print usage
+NOTE: This script should be run as root
+EOT
+}
+
+if [ "$(id -u)" != "0" ]
+then
+    printf "Make sure this script is run as root\\n"
```
oliv3r: i don't really like these enforcement early on. You require your id to be root, which already doesn't _have_ to be true, let the script fail where it fails if you do not have the proper permissions (unless it does not fail due to permission denied of course).

btw, this indicates that the chown (id -u) bit from above also does not work then. IF id -u returns 0 when run under sudo, the chown command above will also return 0 on the chown and again, the file will be owned by 0:0 anyhow. Maybe the environment variable ${USER} (and if it exists ${GROUP} could be used instead then.

build.sh
```diff
@@ -0,0 +1,124 @@
...
+    usage
+    exit
```
oliv3r: this seems like an error case, as such should return exit 1

build.sh
```diff
@@ -0,0 +1,124 @@
...
+fi
+
+while getopts ":hc" options
+do
```
oliv3r: as with the if; then construct

build.sh
```diff
@@ -0,0 +1,124 @@
...
+  case "${options}" in
```
oliv3r: you forgot 2 spaces on your case indent :)

build.sh
```diff
@@ -0,0 +1,124 @@
...
+      exit ${OK}
+      ;;
+    h )
+      usage
+      exit ${OK}
+      ;;
+    :)
+      printf "Option -%s requires an argument.\\n" "${OPTARG}"
+      ;;
+    \?)
```
oliv3r: according to http://pubs.opengroup.org/onlinepubs/9699919799/utilities/getopts.html there is no need to escape the question mark

build.sh
```diff
@@ -0,0 +1,124 @@
...
+  case "${options}" in
+    c )
+      cleanup
+      exit ${OK}
+      ;;
+    h )
+      usage
+      exit ${OK}
+      ;;
+    :)
```
oliv3r: what does this do? i can't find it in the manual page, the manual does say that if the argument is missing, it will be printed out as such on stdarg

I tried to trigger this error but didn't seem to work.

build.sh
```diff
@@ -0,0 +1,124 @@
...
+      printf "Option -%s requires an argument.\\n" "${OPTARG}"
+      ;;
+    \?)
+      printf "Invalid option: -%s\\n" "${OPTARG}"
+      ;;
+  esac
+done
+shift "$((OPTIND-1))"
```
oliv3r: spaces around the subtraction operator

build.sh
```diff
@@ -0,0 +1,124 @@
...
+then
+    printf "Make sure this script is run as root\\n"
+    usage
+    exit
+fi
+
+while getopts ":hc" options
+do
+  case "${options}" in
+    c )
```
oliv3r: to be consistent with your style below, drop the space here and with the 'h' option.

build.sh
```diff
@@ -0,0 +1,124 @@
...
+      ;;
+    \?)
+      printf "Invalid option: -%s\\n" "${OPTARG}"
+      ;;
+  esac
+done
+shift "$((OPTIND-1))"
+
+
+cleanup
```
oliv3r: what is the purpose of this?
If I run the script with -c it will do a cleanup
if I run the script withOUT -c it will do a cleanup?

build.sh
```diff
@@ -0,0 +1,124 @@
...
+    usage
+    exit
+fi
+
+while getopts ":hc" options
+do
+  case "${options}" in
+    c )
+      cleanup
+      exit ${OK}
```
oliv3r: lol i had to look up what 'ok' meant. I think by making it 'more' readable, you made it less readable.
I would leave it at 'exit 0' whoever reads this script will get confused by this, and if someone does not know what exit 0 is, they should not be editing this script anyway.

It also is risky, as if anything modifies the 'OK' parameter, it will now exit that instead, which may be wrong.

Finally, shell scripts only are supposed to have 2 return values anyway, 0 or 1. Though it does work technically to have more, and in that case, 0 is still okay, (always, everywhere and well known) and if it is not 0, use the 'errno' return values (which are not predefined and would need to be defined) see here for a list http://www.tldp.org/LDP/abs/html/exitcodes.html which IS bash specific however.

see here for an interesting (but very basic) blog post http://bencane.com/2014/09/02/understanding-exit-codes-and-how-to-use-them-in-bash-scripts/

build.sh
```diff
@@ -0,0 +1,124 @@
...
+      ;;
+    h )
+      usage
+      exit ${OK}
+      ;;
+    :)
+      printf "Option -%s requires an argument.\\n" "${OPTARG}"
```
oliv3r: assuming the : optind stays, it an exit 1 would be valid needed here, as now it runs the rest of your script without a valid option

build.sh
```diff
@@ -0,0 +1,124 @@
...
+  esac
+done
+shift "$((OPTIND-1))"
+
+
+cleanup
+prepare
+bootstrap_rootfs
+strip_rootfs
+compress_rootfs
```
oliv3r: i like it if scripts that finish successfully explicitly
```sh
...
compress_rootfs

exit 0
```
at the end. no junk (hidden or otherwise) can mess with it then
(echo "rm -rf /" >> $script is always fun)

build.sh
```diff
@@ -0,0 +1,124 @@
...
+      ;;
+    h )
+      usage
+      exit ${OK}
+      ;;
+    :)
+      printf "Option -%s requires an argument.\\n" "${OPTARG}"
+      ;;
+    \?)
+      printf "Invalid option: -%s\\n" "${OPTARG}"
```
oliv3r: missing exit 1

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+
+OK=0
+SRC_ROOT=$(pwd)
+CLEANUP="false"
+RUN_TESTS="false"
+
+usage()
+{
+cat <<-EOT
+Usage:   ./${MY_NAME} [ -h | -c ]
```
oliv3r: ./build_for_ultimaker.sh: 12: ./build_for_ultimaker.sh: MY_NAME: parameter not set
but as before, just "${0}" or "${PROG_NAME}" with [OPTION] ...

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+-c   Cleanup, after a run you have the possibility to get the artifact and then cleanup
+-h   Print usage
+NOTE: This script should be run as root
+EOT
```
oliv3r: why are the command line arguments the same as in the build.sh script?? the build script should just call the build.sh script no?

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+      ;;
+    :)
+      printf "Option -%s requires an argument.\\n" "${OPTARG}"
+      ;;
+    \?)
+      printf "Invalid option: -%s\\n" "${OPTARG}"
+      ;;
+  esac
+done
+shift "$((OPTIND-1))"
```
oliv3r: generally, all remarks as with build.sh

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+NOTE: This script should be run as root
+EOT
+}
+
+if [ "$(id -u)" != "0" ]
+then
+    printf "Make sure this script is run as root\\n"
+    usage
+    exit
+fi
```
oliv3r: as above, though I don't think the build_for_ultimaker.sh script should require any perimssions. that should be contained in build.sh!!

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+  case "${options}" in
+    c )
+      CLEANUP="true"
+      ;;
+    h )
+      usage
+      exit ${OK}
+      ;;
+    t )
+      RUN_TESTS="true"
```
oliv3r: hmm, why do we trigger tests from here?
(from what i remember what you wrote, I think these are checks if all pre-requirements are met)

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/sh
+
+set -eu
+
+OK=0
+SRC_ROOT=$(pwd)
```
oliv3r: this is only true if we actually run this script from the src dir, the jedi-build script for example just calls the scripts afaik.
so then it's just the start_dir or something

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+      printf "Invalid option: -%s\\n" "${OPTARG}"
+      ;;
+  esac
+done
+shift "$((OPTIND-1))"
+
+
+# Check build environment requirements
+cd tests
+./buildenv.sh
```
oliv3r: my first reaction was 'why is the local build script checking the requirements, this is the CI script that needs to check stuff'

Then i realized, this is not a bad idea at all and we should have that (at least as an option as you are doing) in all our stuff.

However, we should only run it per request, not always like here.

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+SRC_ROOT=$(pwd)
+CLEANUP="false"
+RUN_TESTS="false"
+
+usage()
+{
+cat <<-EOT
+Usage:   ./${MY_NAME} [ -h | -c ]
+-c   Cleanup, after a run you have the possibility to get the artifact and then cleanup
+-h   Print usage
```
oliv3r: where did -t go? it's sad and lonely and lost! :hamster:

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+    printf "Make sure this script is run as root\\n"
+    usage
+    exit
+fi
+
+while getopts ":hct" options
+do
+  case "${options}" in
+    c )
+      CLEANUP="true"
```
oliv3r: don't use the string "true" you can just use 'true' which should call /usr/bin/true and return 1 (or whatever true may be)

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+      ;;
+    \?)
+      printf "Invalid option: -%s\\n" "${OPTARG}"
+      ;;
+  esac
+done
+shift "$((OPTIND-1))"
+
+
+# Check build environment requirements
```
oliv3r:
```sh
if ${CHECK_ENV}; then
    env_check()
fi
```
note that i'm not sure what shellcheck says about missing 'glob' quotes here, and if this is actually valid syntax

i think the var equals 0 or 1 here, rather then the command 'true' and 'false' so it may be needed to do if [ 0 ] and also not sure how valid that is

the function is then self responsible for cd-ing to where it wants to be, and returning to the current cwd. from where the function was started. much less error prone

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+cd tests
+./buildenv.sh
+cd "${SRC_ROOT}"
+
+# Build the rootfs image
+./build.sh
```
oliv3r: you are not passing your arguments along "${*}"

also, the local build should perfer to run in docker if available no? you can check the build_for_ultimaker script in the um-kernel repo

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Test the rootfs image
```
oliv3r, line 70: oh huh, ok so i thought the ./buildenv.sh check was the 'test'. right, okay
as mentioned above, having the test run from the buildscript is not a bad idea

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+if [ "${RUN_TESTS}" = "true" ]
```
oliv3r: i guess this works just as well as using /usr/bin/true and probably less of a hassle. :+1:

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+then
+    cd tests
+    ./rootfs.sh -r "${SRC_ROOT}/build/rootfs"
+    cd "${SRC_ROOT}"
```
oliv3r: put it in a function to make it have the same pattern as the env_check() which i still think is nicer (even though this is already very compact)

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+fi
+
+# Cleanup ( this will also delete the rootfs archive from the build dir )
+if [ "${CLEANUP}" = "true" ]
+then
+    ./build.sh -c
+fi
```
oliv3r: remove this entire thing and have -c go to the build-script instead
otherwise, you now have to maintain two interfaces (usage, etc etc) to the same thing.

as for how do i list the different versions:
have a 'build-help' option which does a ./build.sh -h

build_for_ultimaker.sh
oliv3r: maybe put build.sh in a parameter at the start BUILDSCRIPT="${BUILDSCRIPT:-./build.sh}" as it's used more then once in this script

build_for_ultimaker.sh
```diff
@@ -0,0 +1,73 @@
...
+
+exit "${OK}"
```
oliv3r: ah hah! so you do agree and do an exit at the end of your script here :p
but also here, drop the OK; i had to double check the script for occurrences of OK which could have set this, so was not sure what the behavior was anymore.

tests/buildenv.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+set -eu
+
+ERROR=1
```
oliv3r: interesting, a second exit level this may be?

tests/buildenv.sh
```diff
@@ -0,0 +1,26 @@
...
+OK=0
+RESULT="${OK}"
+
+set_error()
+{
+    RESULT="${ERROR}"
+}
+
+are_all_build_env_preconditions_met()
```
oliv3r: ugh, what a function name. very clear, but oh so long

ultimately this is completely up to you, i'm sure there are many books filled with what's better (KISS or descriptive)
preconditions_check() which would do exactly that, check preconditions and exit() on failure.
would be just as well :p

tests/buildenv.sh
```diff
@@ -0,0 +1,26 @@
...
+{
+    command -v debootstrap >/dev/null 2>&1 || { printf >&2 "Build env dependency missing: please install 'debootstrap'\\n"; set_error; }
```
oliv3r: this is very very hard to read. can we not put this in a function so we can use indentation?
```sh
precondition_check()
{
    local CMD="${1}" # I don't think we have to check the existance of $1 as set -u will then fail though in a very simply function like this, just using ${1} isn't a crime either, up to you in this case :)

    command -V "${CMD}" || {{
        echo "please install ... ${CMD}"
        RESULT=1
   }
}

```
further, I think -V is nicer then -v for our purpose (we are not capturing the result to use it, so a nicer output print is good as it informs us WHICH version it found) and thus if things later fail, we know why). Also I ran the script a few times until I ran it with sh -x to find out it was working, but was not giving feedback :) the -V also prints 'not found' if it cannot find it ...

as for set_error() at the end, just set the global RESULT variable to 1. This makes it much more readable. I had to do 3 levels to find what was going on. KISS :)

tests/buildenv.sh
```diff
@@ -0,0 +1,26 @@
...
+    command -v qemu-arm-static >/dev/null 2>&1 || { printf >&2 "Build env dependency missing: please install 'qemu-arm-static'\\n"; set_error; }
+    command -v dd >/dev/null 2>&1 || { printf >&2 "Build env dependency missing: please install 'dd'\\n"; set_error; }
+    command -v mkfs.ext4 >/dev/null 2>&1 || { printf >&2 "Build env dependency missing: please install 'mkfs.ext4'\\n"; set_error; }
+    command -v touch >/dev/null 2>&1 || { printf >&2 "Build env dependency missing: please install 'touch'\\n"; set_error; }
+    command -v chroot >/dev/null 2>&1 || { printf >&2 "Build env dependency missing: please install 'chroot'\\n"; set_error; }
+}
```
oliv3r: i think that this function should
```sh
    command -V .....

    if [ "${RESULT}" -ne 0 ]; then
        exit "${RESULT}" # return "${RESULT}" if you really prefer
    fi
}
```
at its end as there is usually little point in continuing (yeah there is nothing after this function right ow) if we have unmet pre-conditions.

tests/buildenv.sh
```diff
@@ -0,0 +1,26 @@
...
+
+are_all_build_env_preconditions_met
+
+exit "${RESULT}"
```
oliv3r: and then this can be a regular exit 0 again

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+ERROR=1
+OK=0
+MY_NAME=$(basename -- "$0")
+ROOTFS_DIRECTORY="./rootfs"
+TEST_IMAGE_FILE_PATH="${ROOTFS_DIRECTORY}/tmp/test.img"
+TEST_EMPTY_MTAB_FILE="${ROOTFS_DIRECTORY}/etc/mtab"
+
+test_setup()
+{
+    if [ ! -f "${ROOTFS_DIRECTORY}/usr/bin/qemu-arm-static" ]
```
oliv3r: same arguments as above -x and mount --bind

build.sh
oliv3r: also, your usage of command -v "${QEMU_STATIC_BIN}" in the other script below is far more sexy!

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
+#!/bin/sh
+
+set -eu
+
+ERROR=1
+OK=0
+MY_NAME=$(basename -- "$0")
+ROOTFS_DIRECTORY="./rootfs"
+TEST_IMAGE_FILE_PATH="${ROOTFS_DIRECTORY}/tmp/test.img"
```
oliv3r: why not store the test image in a TEST_IMAGE_FILE_PATH="${TEST_IMAGE_FILE_PATH:-$(mktmp)}"

e.g. a temporary random file by default, with override usage if running on a weird system

tests/buildenv.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+set -eu
```
oliv3r, line 3: run test scripts with set -eux by default. it is more verbose, but they are tests after all and improves debug ability (especially on the CI console but even so locally)

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
+#!/bin/sh
+
+set -eu
```
oliv3r, line 3: dito, all tests should use set -eux

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+TEST_EMPTY_MTAB_FILE="${ROOTFS_DIRECTORY}/etc/mtab"
+
+test_setup()
+{
+    if [ ! -f "${ROOTFS_DIRECTORY}/usr/bin/qemu-arm-static" ]
+    then
+        cp "$(command -v qemu-arm-static)" "${ROOTFS_DIRECTORY}/usr/bin/"
+    fi
+
+    dd if=/dev/zero of="${TEST_IMAGE_FILE_PATH}" bs=2M count=64 >/dev/null 2>&1 || return ${ERROR}
```
oliv3r: while we are only creating a very small partition for this test (32 MB?) a block size of 2M is not super ideal.
While there is no 'right' answer here, and bigger is not always better, 2M is relatively small.

I suggest using a single 32M block in this case, which tells us immediately we have a 32mb image :)
secondly, why capture the output; let it go! it's a test after all :) if this fails, we don't even know why, just that test_setup failed.

Finally haven't used return in ages and thus forgot it's true meaning :p (you can't actually return anything, you just set the exit status of the function call) so in that case, the use seems appropriate, but as before, just return 1 in case of error.

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+    touch "${TEST_EMPTY_MTAB_FILE}" >/dev/null 2>&1 || return ${ERROR}
+    mkfs.ext4 "${TEST_IMAGE_FILE_PATH}" >/dev/null 2>&1 || return ${ERROR}
+    mount -o bind /dev/ "${ROOTFS_DIRECTORY}/dev/" >/dev/null 2>&1 || return ${ERROR}
```
oliv3r: i think mount --bind is the modern recommended short for -o bind. ymmv however
anyway, now it becomes tricky. WHY do we need /dev (and not /proc /sys).
Whiping /dev is now very easy (from within the chroot) and i start to worry. On a VM or container this is not a huge deal due to isolation, on my local system running a local test, I get nervous.

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+}
+
+test_destroy()
+{
+    umount -l "${ROOTFS_DIRECTORY}/dev"
```
oliv3r: lazy umount can be a blessing or a curse. Right now, if something is still using /dev it will schedule the umount, but not block on it.
so a rm -rf /dev can still wipe files!

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+        rm -f "${ROOTFS_DIRECTORY}/usr/bin/qemu-arm-static"
+    fi
+}
+
+run_test()
+{
+    if ! test_setup
+    then
+        printf "Cannot run tests, unable to complete test setup\\n"
+        test_destroy
```
oliv3r: i think the common name is test_setup and test_teardown (set UP and tear DOWN) destroy ... don't destroy my system :(

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+        exit ${ERROR}
```
oliv3r: and here we are in a jiffie sadly functions do not return anything, they set exit codes. so how do we get the exit code in the variable without using a global variable. (if we are using a global var, why return anything anyway).

so RESULT=test_setup doesn't exist. what happens if we capture the exit code, e.g. RESULT=$(test_setup) sub-shell execution, i guess it won't try to run a local function .. hmm, while the global var works, it's also ambiguous.
the only thing come up with is
```sh
test_setup
RESULT="${?}"
if [ "${RESULT}" -ne 0 ]; then
    exit "${RESULT}"
fi
```
which is a bit equally ugly, but the lesser of the two evils. imrpovements welcome

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+    fi
+
+    if "$1"
```
oliv3r: hmm, you run the setup and teardown for each test ... it is not overly bad, but can we not use a cascading scheme?

test_setup
runtest1
runtest2
...
test_teardown

would safe on wear and tear on my flash drive :( I don't want to run this on my flash storage for sure :)
if you do prefer to have a setup each time, put the img in a tmpfs/ramdisk instead! unlimited super fast bring-up and teardown. Actually much more preferable!

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+        printf "Run: %s - OK\\n" "${1}"
+    else
+        printf "Run: %s - ERROR\\n" "${1}"
+    fi
+    test_destroy
+}
+
+test_execute_resize2fs()
+{
+    ( chroot "${ROOTFS_DIRECTORY}" /sbin/resize2fs /tmp/test.img >/dev/null 2>&1 && return "${OK}" ) || return "${ERROR}"
```
oliv3r: i agree the tests are simple, but do indent them to improve readability please

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+{
+    ( chroot "${ROOTFS_DIRECTORY}" /sbin/mkfs.f2fs /tmp/test.img >/dev/null 2>&1 && return "${OK}" ) || return "${ERROR}"
+}
+
+usage()
+{
+cat <<-EOT
+Usage:   ./${MY_NAME} -r path to bootstrapped rootfs [ | -h ]
+-h   Print usage
+-r   Path to the bootstrapped rootfs that needs to be tested.
```
oliv3r: since this is a required parameter, it's also an idea to put it at the end as arugment:
`Usage: ${PROG_NAME} [OPTION] ... <bootstrapped_rootfs.img>`
btw, why a bootsrapped fs? are we not creating an empty test image to test the utils?

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+NOTE: This script should be run as root
+EOT
```
oliv3r: as earlier

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+}
+
+if [ "$(id -u)" != "0" ]
+then
+    printf "Make sure this script is run as root\\n"
```
oliv3r: as earlier

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+do
+  case "${options}" in
+    h )
+      usage
+      exit ${OK}
+      ;;
+    r )
+      ROOTFS_DIRECTORY="${OPTARG}"
+      TEST_IMAGE_FILE_PATH="${ROOTFS_DIRECTORY}/tmp/test.img"
+      TEST_EMPTY_MTAB_FILE="${ROOTFS_DIRECTORY}/etc/mtab"
```
oliv3r: this is the final notion of the mtab file, but i don't see it being _used_ anywhere. I suppose some intermediate test fails with this missing? As it can also be empty, can we not bind mount our current system one?

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+    exit "${ERROR}"
+fi
+
+run_test test_execute_resize2fs
+run_test test_execute_fdisk
+run_test test_execute_mount
+run_test test_execute_rsync
+run_test test_execute_busybox
+run_test test_execute_mkfs_ext4
+run_test test_execute_mkfs_f2fs
```
oliv3r:
```sh
...
run_test test_execute_mkfs_f2fs

exit 0
```

tests/rootfs.sh
oliv3r: i guess some checks are needed to ensure we are not umount ////dev
also, make sure that this function _always_ works so that we can call it at_exit setup_teardown EXIT

in case of anything failing, the env is always brought down.

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+
+    if [ -f "${TEST_IMAGE_FILE_PATH}" ]
+    then
+        rm -f ${TEST_IMAGE_FILE_PATH}
+    fi
+
+    if [ -f "${TEST_EMPTY_MTAB_FILE}" ]
+    then
+        rm -f "${TEST_EMPTY_MTAB_FILE}"
+    fi
```
oliv3r: for both the rm -f's a test to ensure we are only delting a subdir! (on the tmp/ramfs :p)

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+}
+
+test_execute_resize2fs()
+{
+    ( chroot "${ROOTFS_DIRECTORY}" /sbin/resize2fs /tmp/test.img >/dev/null 2>&1 && return "${OK}" ) || return "${ERROR}"
+}
+
+test_execute_fdisk()
+{
+    ( chroot "${ROOTFS_DIRECTORY}" /sbin/fdisk -v >/dev/null 2>&1 && return "${OK}" ) || return "${ERROR}"
```
oliv3r: can we do "$(command -v fdisk)" in the subshell instead?
and then fdisk --version (without silencing it) so we know a) it worked and b) which version (looking back at console logs this can be very valuable)
the long opt so it's immediately clear to the reader here, and not silenced so we can see errors and understand why!

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+}
+
+test_execute_mount()
+{
+   ( chroot "${ROOTFS_DIRECTORY}" /bin/mount >/dev/null 2>&1 && return "${OK}" ) || return "${ERROR}"
```
oliv3r: mount --version as above

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+}
+
+test_execute_rsync()
+{
+    ( chroot "${ROOTFS_DIRECTORY}" /usr/bin/rsync -h >/dev/null 2>&1 && return "${OK}" ) || return "${ERROR}"
```
oliv3r: rsync --version

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+}
+
+test_execute_busybox()
+{
+    ( chroot "${ROOTFS_DIRECTORY}" /bin/busybox --help >/dev/null 2>&1 && return "${OK}" ) || return "${ERROR}"
+}
+
+test_execute_mkfs_ext4()
+{
+    ( chroot "${ROOTFS_DIRECTORY}" /sbin/mkfs.ext4 /tmp/test.img >/dev/null 2>&1 && return "${OK}" ) || return "${ERROR}"
```
oliv3r: initially i wanted to write 'mkfs.ext4 -V' so we can get our version info, so maybe do two test here, first if we can find the binary (and which version, for the logs) and then create a simple img.

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+}
+
+test_execute_mkfs_f2fs()
+{
+    ( chroot "${ROOTFS_DIRECTORY}" /sbin/mkfs.f2fs /tmp/test.img >/dev/null 2>&1 && return "${OK}" ) || return "${ERROR}"
```
oliv3r: mkfs.f2fs doesn't need any arguments for the version info :D

tests/rootfs.sh
```diff
@@ -0,0 +1,149 @@
...
+}
+
```
oliv3r: i'm missing the two resize2fs and f2fs.resize tests/tools (which can use the test images, if they not destroyed of course)

oliv3r (changes requested):

tests/buildenv.sh
raysiudak: Clearly not a fan of writing things out.. :-) 
I can argue that this is both descriptive and KISS. Yest it makes the script a few bytes larger and yes if you are working in VGA mode or something it could be more difficult to read. 
To build your case i was more thinking in the line of .. that it would me nice to separate tests in clearly defined functions, then we can add more buildenv test functions later. But YAGNI kicks in and therefore i might as well remove it completely from the function at this point.

tests/buildenv.sh
raysiudak: There is a reason for this. Normally the 1 would resolve in true and 0 in false. Therefore using this in conditional checking improves the readability. You know last time when we where breaking our brains over why the '[' command evaluates '0' as true?

build.sh
raysiudak: 1) An example says more than 1000 words: 

`> echo "$(basename "--filename.sh")"`
`basename: unrecognized option '--filename.sh'`
`Try 'basename --help' for more information.`

`> echo "$(basename "-filename.sh")"`
`basename: invalid option -- 'f'`
`Try 'basename --help' for more information.`

2) Hmm yeah your right, maybe the basename command takes care of this?

build.sh
raysiudak: I think it is useful to keep the build output even after reboot. Maybe that is just the compiled software programmer in me talking? 

Anyway the is_substring check was there because i first had the output dir configurable, but now it is fixed and constant. Therefore this is not needed anyway, we know exactly what we are deleting.

build.sh
raysiudak: 1) Cannot comply: command -V qemu-${ARCH}-static, ARCH = armhf, should be qemu-arm-static. 
And '-V gives more then just the executable path, it should be '-v'.
2) Yes good point

build.sh
raysiudak: this is what prepare_bootstrap does

build.sh
raysiudak: Ok better decide now, so that we do not put more effort in than we have to.

build.sh
raysiudak: As soon as we run the script in a docker container it is not required anymore, there you need to be privileged.

build.sh
raysiudak: Yes this is not needed now since there is no parameter that requires an argument. I always leave the default cases in as in a 'template' so it is easier when adding extra parameters. 
Anyway to enable this for an option you should add ":" after the option definition 'while getopts ":hc:'

build.sh
raysiudak: Yes it will cleanup "before" it builds, to start fresh. After a build we want to be able to fetch the artifact from the build directory and then cleanup after if required.

build.sh
raysiudak: I think this only works on directory's

build.sh
raysiudak: Probably the shellcheck on gitlab is from a different version and it was complaining about it.

build.sh
raysiudak: Sure but do we want a versioned file? If so we should make a new task with requirements for this and implement it there.

build.sh
raysiudak: Debootstrap requires root permissions to run and the idea of this is from the user perspective, they will just run this and then it will build and eventually it will fail somewhere in the end.

tests/rootfs.sh
raysiudak: For this to work i need to create a file first and then mount it. This makes it more complex and also risks of removing the original executable because the script is run as root. I prefer doing it like this, can you agree?

tests/rootfs.sh
raysiudak: Yes removed the whole OK ERROR stuff. 
But 32MB does not work for f2fs it needs > 100MB
Therefore made 4 x 32MB

tests/rootfs.sh
raysiudak: The return statement sets the exit code of the subshell function call, this can be interpreted by the if statement without using the '[' program and therefore works. e.g.: 

> if function; then ....

I do not want to use it in a variable.

tests/rootfs.sh
raysiudak: Yes every test should be run with a clean slate. 
But creating the file in tmp is a good idea

tests/rootfs.sh
raysiudak: command is not part of the rootfs, so this can be used. Anyhow -v tests execution anyway so it is not needed. 
I agree with not silencing stderr, but not for stdout.

tests/rootfs.sh
raysiudak: Id rather test if it actually formats

tests/rootfs.sh
raysiudak: resize2f is one of the first tests, i think you missed it.

tests/rootfs.sh
raysiudak: No we are testing a bootstrapped rootfs created with the build.sh script

tests/rootfs.sh
raysiudak: Yess  both mkfs.f2fs and resize2fs require the existence of this file. 
Bind mounting is a good idea so that the tools can check if the file is mounted, but then it is not required for the test because they succeed anyway and it requires more loc. 
But then again, maybe this is an error of the tool and therefore in a future version this will be fixed, so gonno bind mount it anyway.

tests/rootfs.sh
raysiudak: debootstrap requires dev mount to run, but it can be mounted read-only if that makes you feel better :-P

build.sh
raysiudak: No when bind mounting the file should exist and therefore it would not be more simple, it would make it more dangerous because something might happen to the original file.

build_for_ultimaker.sh
raysiudak: already renamed to CUR_DIR

build_for_ultimaker.sh
raysiudak: aaahhh so sad, build script should not have knowledge of the test script.

build_for_ultimaker.sh
raysiudak: No before the build start, the preconditions are checked, then the rootfs is build and you can choose to run the tests on the rootfs.

build_for_ultimaker.sh
raysiudak: I am not doing this because 'build_for_ultimaker.sh' is for local usage and 'build.sh' is to be used in this script or by docker.

raysiudak: no its not

build_for_ultimaker.sh
raysiudak: /usr/bin/true as you said could not exist or contain something else, so renamed it to yes and no, its a little better

build_for_ultimaker.sh
raysiudak: Removed it because when building it clean before so it does not make sense anyway. The clean is used by docker.

build.sh
oliv3r:
```sh
> export FN=build_for_ultimaker.sh
> echo "$(basename "${FN}")"
build_for_ultimaker.sh
```
(the export here is needed to export the variable to the (sub)shell when executing from the cli)

build.sh
oliv3r: :+1: you are right, we do not want this to go away after a reboot!

build.sh
```diff
@@ -0,0 +1,118 @@
...
+
+
+cleanup()
+{
+    local mount_points
+    mount_points="$(mount | grep "${ROOTFS_DIR}")"
+
+    if [ "${mount_points}" = "" ]; then
+        if [ -d "${BUILD_DIR}" ]; then
+            rm -rf "${BUILD_DIR}"
```
oliv3r: your core step of the cleanup step is deeply nested and hidden, personally i prefer to avoid nesting by fail often and fail early.

build.sh
```diff
@@ -0,0 +1,118 @@
...
+DEBIAN_VERSION="${DEBIAN_VERSION:-jessie}"
+ROOTFS_ARCHIVE="rootfs.tar.xz"
+ROOTFS_DIR="${BUILD_DIR}/rootfs"
+QEMU_STATIC_BIN="$(command -v qemu-arm-static)"
+
+
+cleanup()
+{
+    local mount_points
+    mount_points="$(mount | grep "${ROOTFS_DIR}")"
```
oliv3r: does this actually work?
the return value of grep if it does not find anything is 1 (error) because one can use the return code of grep to check if something occurs in a file.

in other words, grep fails to find anything, the script stops.
we save ourselves a little trouble by asking the kernel directly, e.g. grepping /proc/mounts directly
```sh
grep "${ROOTFS_DIR}" /proc/mounts || true
```
edit: after testing, it turns out, it doesn't work, the script just exists at the cleanup stage

build.sh
oliv3r: is that how it's going to be :p
yeah, -v is what we want here, as -V is only useful for the 'is it available' check where it prints it nicely and verbose. My mistake.
Also I did not name the binary correctly, there's an annoyance right now, that there is not a single name for ARM currently.
There is:
arm, armel, armhf, aarch64
some scripts/vars want armhf, some are fine with arm.
The kernel only accepts 'ARCH=arm' whereas debootstrap wants armhf :S

for now this will work as we only have armhf; this may (though i think it actually works) fail on aarch64 and will fail on mips :)
the only solution that I have come up with so far, is to do something ugly like:
```sh
if [ $ARCH -eq "arm" ]; then ARCH="armhf" fi
```

build.sh
```diff
@@ -0,0 +1,118 @@
...
+
+    if [ ! -x "${ROOTFS_DIR}${QEMU_STATIC_BIN}" ]; then
+        cp "${QEMU_STATIC_BIN}" "${ROOTFS_DIR}/usr/bin/"
+    fi
+}
+
+un_prepare_bootstrap()
+{
+    if [ -f "${ROOTFS_DIR}${QEMU_STATIC_BIN}" ]; then
+        rm  -f "${ROOTFS_DIR}${QEMU_STATIC_BIN}"
```
oliv3r: over eager on the whitespace there :p

build.sh
```diff
@@ -8,6 +8,7 @@ BUILD_DIR="${CUR_DIR}/.build_${ARCH}"
 DEBIAN_VERSION="${DEBIAN_VERSION:-jessie}"
```
oliv3r: abusing this spot for a small note;
don't make commit messages in the past tence 'e.g. Make use of' rather then 'made use of'. This commit performs an action, it will do it when applying the commit.
Its just semantics so don't worry about changing it, just something for the future to keep in mind

build.sh
```diff
@@ -40,14 +41,14 @@ prepare_bootstrap()
 un_prepare_bootstrap()
 {
     if [ -f "${ROOTFS_DIR}${QEMU_STATIC_BIN}" ]; then
-        rm  -f "${ROOTFS_DIR}${QEMU_STATIC_BIN}"
+        "${RUN_SUDO}" rm  -f "${ROOTFS_DIR}${QEMU_STATIC_BIN}"
```
oliv3r: I'm wondering, why do we need sudo to remove the file, but not when copying it to the location?
I suspect copy operation will place the file in the rootfs (as $USER) and thus we should be able to delete it again as well? It is our file after all.

build.sh
oliv3r: It can be done actually. Docker does this, and i tried (and failed it) after searching it, it is possible.
https://unix.stackexchange.com/questions/44373/single-bind-mounted-file-gets-out-of-sync-in-linux#44375

the trick is to do:
```sh
> mount --bind /usr/bin/qemu-arm-static /tmp/qemu-arm-static                                                                                                                                            mount: /tmp/qemu-arm-static: mount point does not exist.
> touch /tmp/qemu-arm-static                                                                                                                                                                                > sudo mount --bind /usr/bin/qemu-arm-static /tmp/qemu-arm-static                                                                                                                                           
-rw-r--r-- 1 oliver oliver 0 jul 24 12:30 /tmp/qemu-arm-static                                                                                                                                                                             > sudo umount /tmp/qemu-arm-static    
> rm /tmp/qemu-arm-static                                                                                                                                                                                                                                             
```
Can't decide which approach i find cleaner, as we still have to create a placeholder file, mount it, and remove the placeholder file. I really like how it works in docker (and docker does this transparently)
and since we cannot do this with a mkdir, but only with a placeholder file (touch) the code won't become much simpler?
thoughts?

build.sh
oliv3r: well what do you think :p though it would only change this one function so the change is rather minimal

```sh
    mkdir -p "${ROOTFS_DIR}/etc/apk"
    echo "${ALPINE_REPO}/${ALPINE_VERSION}/main" > "${ROOTFS_DIR}/etc/apk/repositories"
    # Install rootfs with base applications
    apk --root "${ROOTFS_DIR}" --update-cache \
       add --allow-untrusted --initdb --arch "${ARCH}" \
       busybox e2fsprogs f2fs-tools rsync
    # Add baselayout etc files
    apk --root "${ROOTFS_DIR}" \
       fetch --arch "${ARCH}" --stdout alpine-base | tar -xvz -C "${ROOTFS_DIR}" etc
    rm -f "${ROOTFS_DIR}/var/cache/apk"/*
```
from https://github.com/gliderlabs/docker-alpine/blob/rootfs/library-3.6/armhf/builder/scripts/mkimage-alpine.bash#L28

this yields an image that's about 1.3MB!! (sadly still to large for our current initramfs, but we are getting somewhere!

the only downside is, that running this locally is harder (needs apk installed locally) when not running inside a docker container (which in turn is easier). Granted, right now we require lots of tools to be installed locally, so this is not something holding us back.

we could even crazy and make dockerfile, and docker docker build && docker export && docker rm :p

raysiudak: Sure let's decide after we do the stripping

tests/buildenv.sh
oliv3r: yeah, test ('[') considers the return value of 0 to be 'ok' e.g. true.
But this is a codeing style/practise change, that has very deep inpakt and involves risk.
People, other then you and me, will be unfamiliar with this construct, and what if someone 'abuses' this as 'oh this is the return status, ERROR=0, there is no error here'

even after knowing this, and re-reading your code, i was puzzling 'what is ERROR=1, what does that mean, is that the initial value of the return value?
e.g. a common pattern
```c
int err = 0;

err = somefunc();

return err;
```
is quite common, hence the double confusion. So while result=ERROR; makes makes sense, for readability, it also leads to lots of confusion for people not familiar with this paradigm

tests/buildenv.sh
oliv3r: my dilexia doesn't go well with writing things out too long :) also human brains don't like long things generally :p but yeah, the name doesn't bother me THAT much, I was just curious :)

build.sh
oliv3r: just an extra info; i said that, because your function was doing a boolean check and thus i expect true or false to be returned (and thus checked against), as 0 is normally 'ok' what would the meaning of 0 be in this case? anyway, the function is gone, it does not matter :p

build.sh
```diff
@@ -0,0 +1,118 @@
...
+BUILD_DIR="${CUR_DIR}/.build_${ARCH}"
+DEBIAN_VERSION="${DEBIAN_VERSION:-jessie}"
+ROOTFS_ARCHIVE="rootfs.tar.xz"
+ROOTFS_DIR="${BUILD_DIR}/rootfs"
+QEMU_STATIC_BIN="$(command -v qemu-arm-static)"
+
+
+cleanup()
+{
+    local mount_points
```
oliv3r: sadly, local is not posix compliant :S

build.sh
```diff
@@ -0,0 +1,118 @@
...
+    mount_points="$(mount | grep "${ROOTFS_DIR}")"
+
+    if [ "${mount_points}" = "" ]; then
```
oliv3r: test ('[') has -n and -z because it is different from "". just a note, to watch out for that.

build.sh
oliv3r: what i meant was (which you've added) we must ensure qemu-arm-static is available. you do that via command -V now, so that's good :+1: 
further more, by switching to alpine, we don't need qemu anymore anyway :)

build.sh
oliv3r: i checked, stripping debian won't get us anywhere near the 1.3mb. the smallest debian immages are still 60-ish MB. the current alpine image is 1.3mb which is ridiculously small. it also makes the script much simpler. i'll push my stuff as commits which you can decide what to do with :)

build.sh
oliv3r: all our packages are versioned now, so I would expect this to be the case for this package as well.
We can start initially by doing something like:
VERSION=git describe ...
rootfs-VERSION.tar.xz

build.sh
oliv3r: we perform this check too early. now, it gets run before getops gets run, and as such, we cannot even do getopts -h (without error) as we always trigger the non-root with error case

build.sh
```diff
@@ -0,0 +1,118 @@
...
+
+
+cleanup
+prepare_bootstrap
+bootstrap_rootfs
+un_prepare_bootstrap
+#strip_rootfs
+compress_rootfs
+
+exit 0
```
oliv3r: broken editor

build.sh
oliv3r: nah, read-only mount :)

PR #4, opened 2018-07-30
Made local build script docker compatible

raysiudak:
This means that when docker is installed the script is run in the given docker container, if not it will try locally.
Before the build or the test, the build environment preconditions are checked.

Constributes to: EMP-271, implements EMP-350

### DIFF
--- a/build_for_ultimaker.sh
+++ b/build_for_ultimaker.sh
@@ -2,32 +2,69 @@
 
 set -eu
 
+CI_REGISTRY_IMAGE="${CI_REGISTRY_IMAGE:-registry.gitlab.com/ultimaker/embedded/platform/um-updatefs_armhf}"
+CI_REGISTRY_IMAGE_TAG="${CI_REGISTRY_IMAGE_TAG:-latest}"
+
 ARCH="${ARCH:-armhf}"
 CUR_DIR=$(pwd)
 BUILD_DIR="${CUR_DIR}/.build_${ARCH}"
-ROOTFS_IMG="${BUILD_DIR}/rootfs*.xz.img"
-RUN_ENV_CHECK="no"
-RUN_TESTS="no"
+ROOTFS_IMG="rootfs.xz.img"
+
+DOCKER_WORK_DIR="${DOCKER_WORK_DIR:-/build}"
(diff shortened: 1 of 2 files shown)


build_for_ultimaker.sh
```diff
@@ -2,32 +2,69 @@
 
 set -eu
 
+CI_REGISTRY_IMAGE="${CI_REGISTRY_IMAGE:-registry.gitlab.com/ultimaker/embedded/platform/um-updatefs_armhf}"
```
oliv3r: um-update_toolbox_armhf

build_for_ultimaker.sh
```diff
@@ -2,32 +2,69 @@
...
 }
 
 usage()
 {
 cat <<-EOT
     Usage: ${0} [OPTIONS]
-        -c   Run build environment checks
+        -c   Run build environment checks (run by default when building and testing)
         -h   Print usage
         -t   Run rootfs tests
```
oliv3r: is this not run by default? if not why not?

build_for_ultimaker.sh
```diff
@@ -2,32 +2,69 @@
...
+}
+
+run_build()
+{
+    env_check
+    if command -V docker; then
+        run_in_docker "${DOCKER_WORK_DIR}" "./build.sh" ""
+    else
+        ./build.sh
+    fi
```
oliv3r, line 49: why don't we test by default as well?

build_for_ultimaker.sh
```diff
@@ -56,19 +93,20 @@ while getopts ":cht" options; do
...
 # Check build environment requirements
-if [ "${RUN_ENV_CHECK}" = "yes" ]; then
+if [ "${run_env_check}" = "yes" ]; then
     env_check
+    exit 0
 fi
 
 # Build the rootfs image
-./build.sh
+run_build
```
oliv3r: i would put this below the run test.

run env and run test are run explicitly (without anything else) when requested. when doing a build, we always run the test and env check.

oliv3r (changes requested):

oliv3r (approved):

PR #6, opened 2018-08-02
EMP-356 bind mount qemu arm

oliv3r:


tests/buildenv.sh
```diff
@@ -9,25 +9,27 @@ result=0
 
 check_filesystem_support()
 {
-    fs="${1}"
-    grep "${fs}" "/proc/filesystems" || result=1
+    for fs in ${FILESYSTEMS}; do
```
raysiudak, line 74: Might as well make the FILESYSTEMS constant part of the function then

tests/buildenv.sh
```diff
@@ -9,25 +9,27 @@ result=0
...
+        fi
+
+    done
 }
 
 check_package_installation()
 {
-    cmd="${1}"
-    command -V "${cmd}" || result=1
+    for pkg in ${PACKAGES}; do
```
raysiudak, line 87: Same goes for PACKAGES

tests/buildenv.sh
```diff
@@ -2,12 +2,67 @@
...
+BINFMT_MISC="${BINFMT_MISC:-/proc/sys/fs/binfmt_misc/}"
+PACKAGES="apk mksquashfs xz"
 FILESYSTEMS="overlay squashfs tmpfs"
 
+ARMv7_MAGIC="7f454c4601010100000000000000000002002800"
+
 result=0
 
 
+check_emulation_support()
```
raysiudak: This function is too long and is difficult to understand, could split it up into several self explaining helper functions to improve readability.

tests/buildenv.sh
```diff
@@ -2,12 +2,67 @@
...
+            continue
+        fi
+
+        if grep -q "${ARMv7_MAGIC}" "${emu}"; then
+            __ARM_EMU_BIN="$(grep "interpreter" "${emu}")"
+            __ARM_EMU_BIN="${__ARM_EMU_BIN#interpreter }"
+            break
+        fi
+    done
+
```
raysiudak, line 44: remove one newline

tests/rootfs.sh
```diff
@@ -42,6 +48,9 @@ teardown()
         return
     fi
 
+    umount "${rootfs_dir}/${ARM_EMU_BIN}"
+    unlink "${rootfs_dir}/${ARM_EMU_BIN}"
```
raysiudak, line 52: is rm not sufficient here?

tests/buildenv.sh
oliv3r: We could yes, but you want to have the configurables/main parameters of the script grouped at the top so they are easy to find.

tests/buildenv.sh
oliv3r: :+1:

tests/rootfs.sh
oliv3r: rm > unlink :p so technically every-time you do an rm, you should wonder, is unlink not sufficient here :)

also, purely coincidentally it aligns nicely :p

raysiudak (dismissed):

raysiudak (approved):

PR #7, opened 2018-08-02
EMP-356 bind mount qemu arm

oliv3r:
We merged this before the build could fail/succeed. This fixes our CI failure.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -43,8 +43,16 @@ bootstrap_prepare()
 
 bootstrap_unprepare()
 {
-    umount "${ROOTFS_DIR}/${ARM_EMU_BIN}"
-    unlink "${ROOTFS_DIR}/${ARM_EMU_BIN}"
+    if [ ! -e "${ROOTFS_DIR}/${ARM_EMU_BIN}" ]; then
+        return
+    fi
+
+    if grep -q "$(realpath "${ROOTFS_DIR}/${ARM_EMU_BIN}")" "/proc/mounts"; then
+        umount "${ROOTFS_DIR}/${ARM_EMU_BIN}"
+    fi
+    if [ -f "${ROOTFS_DIR}/${ARM_EMU_BIN}" ]; then
+        unlink "${ROOTFS_DIR}/${ARM_EMU_BIN}"
+    fi
 }
 
 bootstrap_rootfs()
--- a/build_for_ultimaker.sh
+++ b/build_for_ultimaker.sh
@@ -33,8 +33,7 @@ setup_emulation_support()
         fi
 
         if grep -q "${ARMv7_MAGIC}" "${emu}"; then
-            ARM_EMU_BIN="$(grep "interpreter" "${emu}")"
-            ARM_EMU_BIN="${ARM_EMU_BIN#interpreter }"
+            ARM_EMU_BIN="$(sed 's/interpreter //;t;d' "${emu}")"
             break
         fi
     done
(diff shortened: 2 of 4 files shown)


build.sh
```diff
@@ -30,21 +30,21 @@ cleanup()
 
 bootstrap_prepare()
 {
-    if [ ! -x "${ARM_EMU_BIN}" ]; then
+    if [ ! -x "${ARM_EMU_BIN}" ] || [ "${ARM_EMU_BIN#/}" != "${ARM_EMU_BIN}" ]; then
```
raysiudak: In the alternate scenario.. when true, the binary is not checked for execution anymore. 
Maybe it is better to check for execution after the mount?

build_for_ultimaker.sh
```diff
@@ -39,7 +39,7 @@ setup_emulation_support()
         fi
     done
 
-    if [ ! -x "${ARM_EMU_BIN}" ]; then
+    if [ ! -x "${ARM_EMU_BIN}" ] && [ "${ARM_EMU_BIN#/}" != "${ARM_EMU_BIN}" ]; then
```
raysiudak: I am confused.. If you cannot execute the arm binary, and the arm binary is not equal to the arm binary without preceding slash.. then error. 
In this script the ARM_EMU_BIN should always have the full path including the starting '/'

build_for_ultimaker.sh
```diff
@@ -39,7 +39,7 @@ setup_emulation_support()
         fi
```
raysiudak, line 38: By the way, comment on your previous pull request:

grep interpreter "/proc/sys/fs/binfmt_misc/qemu-arm" | sed 's/interpreter //'

tests/buildenv.sh
```diff
@@ -54,7 +54,7 @@ check_emulation_support()
         return
     fi
 
-    if [ ! -x "${ARM_EMU_BIN}" ]; then
+    if [ ! -x "${ARM_EMU_BIN}" ] || [ "${ARM_EMU_BIN#/}" != "${ARM_EMU_BIN}" ]; then
```
raysiudak: Again it is very confusing what actually is going on here, what are your actually testing

tests/rootfs.sh
```diff
@@ -12,7 +12,7 @@ RESULT=0
 
 setup()
 {
-    if [ ! -x "${ARM_EMU_BIN}" ]; then
+    if [ ! -x "${ARM_EMU_BIN}" ] || [ "${ARM_EMU_BIN#/}" != "${ARM_EMU_BIN}" ]; then
```
raysiudak: Also it is nowhere stated in the usage of these script when and what to pass as env with the scripts

raysiudak (changes requested):

build_for_ultimaker.sh
oliv3r: excellent, that's almost the one liner i was looking for and sed is part of busybox.
`
sed 's/interpreter //;t;d' /proc/sys/fs/binfmt_misc/qemu-arm
`
is the command i was looking for!

You are awesome ;)

build_for_ultimaker.sh
oliv3r: exactly, the ARM_EMU_BIN should always be an absolute path, thus always start with a /. This should be the default anyway.

The alternative is to add the / to the ROOTFS_DIR. which after sleeping over it, I think is better, albeit a bigger patch. Then we can support both absolute and relative paths.

raysiudak (approved):

PR #8, opened 2018-08-03
EMP-357 mount toolbox from ramdisk

raysiudak:


.gitlab-ci.yml
```diff
@@ -101,7 +101,7 @@ build package:
   artifacts:
     name: "$CI_COMMIT_REF_NAME-$CI_COMMIT_SHA"
     paths:
-    - ./.build_*/rootfs*.xz.img
+    - ./.build_*/um-update*.xz.img
```
oliv3r, line 104: :+1:

build.sh
```diff
@@ -8,7 +8,7 @@ CUR_DIR="$(pwd)"
 BUILD_DIR="${CUR_DIR}/.build_${ARCH}"
 ALPINE_VERSION="${ALPINE_VERSION:-latest-stable}"
 ALPINE_REPO="${ALPINE_REPO:-http://dl-cdn.alpinelinux.org/alpine}"
-ROOTFS_ARCHIVE="rootfs.xz.img"
+TOOLBOX_IMAGE="um-update_toolbox.xz.img"
```
oliv3r: note to ourselves: when we introduce version numbering, we will likely add that to the image name as well, at which point it makes sense to use
```sh
TOOLBOX_IMAGE="${TOOLBOX_IMAGE}:-um-update_toolbox.xz.img"
```

build_for_ultimaker.sh
```diff
@@ -9,7 +9,7 @@ ARCH="${ARCH:-armhf}"
 ARM_EMU_BIN=
 CUR_DIR=$(pwd)
 BUILD_DIR="${CUR_DIR}/.build_${ARCH}"
-ROOTFS_IMG="rootfs.xz.img"
+TOOLBOX_IMAGE="um-update_toolbox.xz.img"
```
oliv3r, line 12: i didn't realize this until now, but this is really ugly, we now have the name in 2 places and they have to match.

In that case, we want to do the proposed above change, and pass this variable to the test script along, so we don't have to keep these names in sync. See EMP-357.

oliv3r (changes requested):

tests/buildenv.sh
```diff
@@ -2,13 +2,13 @@
...
 
+ARMv7_MAGIC="7f454c4601010100000000000000000002002800"
 ARM_EMU_BIN="${ARM_EMU_BIN:-}"
 BINFMT_MISC="${BINFMT_MISC:-/proc/sys/fs/binfmt_misc/}"
+
 PACKAGES="apk mksquashfs xz"
 FILESYSTEMS="overlay squashfs tmpfs"
 
-ARMv7_MAGIC="7f454c4601010100000000000000000002002800"
-
```
oliv3r: :-1: 
why?
ARMv7_MAGIC is an internal non-settable constant, so it is below the 'argument' variables and below the user settable constants.

What IS wrong however is that PACKAGES should be below FILESYSTEMS :p

build.sh
```diff
@@ -55,6 +54,25 @@ bootstrap_unprepare()
     fi
 }
 
+add_scripts()
```
oliv3r: what scripts?
add_update_scripts() maybe?

scripts/startup.sh
```diff
@@ -0,0 +1,6 @@
+# This script is a placeholder that is executed as entry point for system configuration routines.
+
+#!/bin/sh
```
oliv3r: swap comment and the shebang, the shebang should be the first line of a script

tests/toolbox_image.sh
```diff
@@ -129,6 +130,14 @@ test_execute_rsync()
     ( chroot "${rootfs_dir}" /usr/bin/rsync --version 1> /dev/null && return 0 ) || return 1
 }
 
+test_system_update_entrypoint()
+{
+    if [ ! -x "${rootfs_dir}${SYSTEM_UPDATE_ENTRYPOINT}" ]; then
+        return 1
+    fi
+    return 0
```
oliv3r: just return the result of the test
where it may be more readable to use the binary as 'test' rather then the symlink [ (yes that's the name of the binary ;) )
```sh
return test -x "${rootfs_dir}${SYSTEM_UPDATE_ENTRYPOINT}"
```
doesn't even need to be negated :)

build_for_ultimaker.sh
```diff
@@ -97,7 +97,6 @@ cat <<-EOT
         -c   Skip run of build environment checks
         -h   Print usage
         -t   Skip run of rootfs tests
-    NOTE: This script requires root permissions to run.
```
oliv3r, line 100: :+1:

oliv3r (changes requested):

build.sh
```diff
@@ -55,6 +54,25 @@ bootstrap_unprepare()
...
+{
+    local_script_dir="${CUR_DIR}/scripts"
+    target_script_dir="${ROOTFS_DIR}/sbin"
+    entrypoint="startup.sh"
+
+    if [ ! -f "${local_script_dir}/${entrypoint}" ]; then
```
oliv3r, line 63: use -x; the entrypoint script needs to be executable to be able to function as entrypoint

build.sh
```diff
@@ -55,6 +54,25 @@ bootstrap_unprepare()
...
+        echo "Missing entrypoint script '${local_script_dir}/${entrypoint}'"
+        exit 1
+    fi
+
+    for script in ${local_script_dir}/*; do
```
oliv3r: as the linter suggested:
```sh
 for script in "${local_script_dir}/"*; do
```
Note that that if you quote the asterix as well, assuming that globbing works, it would still pass everything as one big string as $script (with spaces) as no string splitting occurs anymore

Note to ourselves, make a simple pre-commit hook that runs the shellchecker. I run it locally already as:
```sh
docker run --rm -v "$(pwd)":"$(pwd)" registry.hub.docker.com/koalaman/shellcheck:stable -Calways -s sh "$(pwd)"/test/rootfs.sh
```
(but also tend to forget it quite often) by documenting this as a standard pre-commit hook for shell scripts or something, we can do the test before the CI has to trigger on it.

build.sh
```diff
@@ -55,6 +54,25 @@ bootstrap_unprepare()
...
+        basename="${script##*/}"
+        echo "Installing ${script} on '${target_script_dir}/${basename}'"
```
oliv3r: if you start a sentence with a capital, end it with a period :p either use a small i, or end with a . :)

oliv3r (changes requested):

build.sh
oliv3r: :crying_cat_face: :hankey:

build_for_ultimaker.sh
raysiudak: Yes it is, it is even more difficult because the init.sh script is also using the image name. In the future we should have some digital signing and get rid of the fixed names.

oliv3r (approved):

PR #15, opened 2018-08-15
EMP-304 dummy storage device

raysiudak:


tests/buildenv.sh
```diff
@@ -5,8 +5,8 @@ set -eu
 ARM_EMU_BIN="${ARM_EMU_BIN:-}"
 BINFMT_MISC="${BINFMT_MISC:-/proc/sys/fs/binfmt_misc/}"
 
-FILESYSTEMS="overlay squashfs tmpfs"
-COMMANDS="apk mksquashfs xz"
+FILESYSTEMS="overlay squashfs tmpfs ext4 f2fs"
+COMMANDS="apk mksquashfs xz sed"
```
oliv3r: keep these variables sorted please

tests/buildenv.sh
```diff
@@ -38,6 +38,7 @@ check_emulation_support()
 
         if grep -q "${ARMv7_MAGIC}" "${emu}"; then
             __ARM_EMU_BIN="$(sed 's/interpreter //;t;d' "${emu}")"
+            echo "${__ARM_EMU_BIN}"
```
oliv3r: this is printed too early. This is only the expected emulator. There are 3 tests following this one which can fail. resulting in a printout saying 'emu support: qemu-arm-static'  'Error, cannot continue'

Right now we print 'ok' whenever all checks have passed. I think it's not a bad idea to print more info on success to yield more information, but then it should be at the end, and in this specific case using ${ARM_EMU_BIN}.

so pick one:
```sh
echo "ok"
```
(that's what we have now)

or
```sh
echo "ok (${QEMU_ARM_BIN)"
```
or something like that.
```

File: tests/toolbox_image.sh [hunk 3]
```diff
@@ -3,8 +3,19 @@
...
 
 ARM_EMU_BIN="${ARM_EMU_BIN:-}"
-TMP_TEST_IMAGE_FILE="/tmp/test_file.img"
+
 SYSTEM_UPDATE_ENTRYPOINT="/sbin/startup.sh"
+DISK_PREPARE_COMMAND="/sbin/prepare_disk.sh"
+
+# Test storage device parameters
+STORAGE_DEVICE_FILE="/tmp/storage_device.img"
+SECTOR_SIZE="512"
```
oliv3r: :+1:

tests/toolbox_image.sh
```diff
@@ -3,8 +3,19 @@
...
+SECTOR_TOTAL="7553024"
```
oliv3r: DEVICE_TOTAL_SIZE (maybe add a comment 'Storage size in sectors, real size xxx GiB)?

tests/toolbox_image.sh
```diff
@@ -3,8 +3,19 @@
...
+BOOT_START="2048"
+ROOTFS_START="67584"
+SPOOL_START="1998848"
```
oliv3r: it is actually the USERDATA_START. We just mount the cluster spool there because we didn't have time to properly implement it, so lets not let this technical debt accumulate further.

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
...
 exit_on_failure=false
 
+format_dummy_partition()
+{
+    command="${1}"
+    sector_start="${2}"
+    device_file="${3}"
+
+    loop_device="$(chroot "${rootfs_dir}" /sbin/losetup -f)"
+    echo "Using next available loop device '${loop_device}' to format partition using: '${command}':"
```
oliv3r: i think you intended a period at the end of this line?

Ah I think it prints the mount point of the if grep at the end here? If so, i think you need an extra space :) (or maybe even ''${command}' on: "?)

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
...
+
+    touch "${rootfs_dir}${loop_device}"
+    mount --bind "${loop_device}" "${rootfs_dir}${loop_device}"
```
oliv3r: i thought you said you wanted to do this during setup/teardown? if you do it in the setup/teardown you can just dd the file INTO the '${rootfsdir}', unless of course you want the image the remain the same between setup-teardowns of course, which makes sense here, as you prepare the image for the last test; though then it does make sense to remove the earlier instance of the test image that is used during setup/teardown :)

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
 
 exit_on_failure=false
 
+format_dummy_partition()
+{
+    command="${1}"
+    sector_start="${2}"
+    device_file="${3}"
+
+    loop_device="$(chroot "${rootfs_dir}" /sbin/losetup -f)"
```
oliv3r: if we chroot, do we not have access to the PATH? I think it's better not to prefix the commands if we don't have to, so that if they'd move, things don't break

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
...
+    mount --bind "${loop_device}" "${rootfs_dir}${loop_device}"
+
+    chroot "${rootfs_dir}" /sbin/losetup -o "${sector_start}" "${loop_device}" "${device_file}"
+    chroot "${rootfs_dir}" "${command}" "${loop_device}"
+    chroot "${rootfs_dir}" /sbin/losetup -d "${loop_device}"
+
+    if grep "${loop_device}" "/proc/mounts"; then
+        umount "${rootfs_dir}${loop_device}"
+        unlink "${rootfs_dir}${loop_device}"
+    fi
```
oliv3r: For concistency with teardown:
```sh
        if [ -f "${loop_device}" ]; then
            unlink "${loop_device}"
        fi
```
so that it uses the same pattern

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
...
+}
+
+create_dummy_storage_device()
+{
+    partition_table_file="${rootfs_dir}/tmp/partition_table"
+
+cat <<EOT >> "${partition_table_file}"
```
oliv3r: you can, if you want, indent cat by one as you expect, and if you use <<-EOT (added hyphen) you can double indent the partition table, as the hyphen removes the indentation in the temporary file) (sadly however, this does not apply to the EOT marker as far as i'm aware.

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
...
+unit: sectors
+
+boot    : start=${BOOT_START},      size=$(( ROOTFS_START - BOOT_START )),    Id=83
+rootfs  : start=${ROOTFS_START},    size=$(( SPOOL_START - ROOTFS_START )),   Id=83
+spool   : start=${SPOOL_START},     size=$(( SECTOR_TOTAL - SPOOL_START )),   Id=83
+EOT
+
+    echo "Creating test image: '${rootfs_dir}${STORAGE_DEVICE_FILE}'"
```
oliv3r: When you start with a capital, end with a period :p

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
...
+}
+
+create_dummy_storage_device()
+{
+    partition_table_file="${rootfs_dir}/tmp/partition_table"
+
+cat <<EOT >> "${partition_table_file}"
+unit: sectors
+
+boot    : start=${BOOT_START},      size=$(( ROOTFS_START - BOOT_START )),    Id=83
```
oliv3r: we only space around curly braces when we initialize arrays

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
...
+spool   : start=${SPOOL_START},     size=$(( SECTOR_TOTAL - SPOOL_START )),   Id=83
+EOT
+
+    echo "Creating test image: '${rootfs_dir}${STORAGE_DEVICE_FILE}'"
+    dd if="/dev/zero" of="${rootfs_dir}${STORAGE_DEVICE_FILE}" bs="1" count="0" \
+        seek="$(( SECTOR_SIZE * SECTOR_TOTAL ))" || return 1
+
+    echo "Writing partition table:"
+    cat "${partition_table_file}"
+    chroot "${rootfs_dir}" /sbin/sfdisk "${STORAGE_DEVICE_FILE}" < "${partition_table_file}" || return 1
```
oliv3r: as with losetup, drop the path if you can

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
...
+unit: sectors
+
+boot    : start=${BOOT_START},      size=$(( ROOTFS_START - BOOT_START )),    Id=83
+rootfs  : start=${ROOTFS_START},    size=$(( SPOOL_START - ROOTFS_START )),   Id=83
+spool   : start=${SPOOL_START},     size=$(( SECTOR_TOTAL - SPOOL_START )),   Id=83
+EOT
+
+    echo "Creating test image: '${rootfs_dir}${STORAGE_DEVICE_FILE}'"
+    dd if="/dev/zero" of="${rootfs_dir}${STORAGE_DEVICE_FILE}" bs="1" count="0" \
+        seek="$(( SECTOR_SIZE * SECTOR_TOTAL ))" || return 1
```
oliv3r: if dd fails, it will return with 1 already, no need to explicitly return on failure

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
 
 exit_on_failure=false
 
+format_dummy_partition()
+{
+    command="${1}"
+    sector_start="${2}"
+    device_file="${3}"
```
oliv3r: device_file == STORAGE_DEVICE_FILE? do we forsee (YAGNI) that we need this parameter variable? It's a 'constant' throughout the program already anyway ...

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
...
+
+    echo "Writing partition table:"
+    cat "${partition_table_file}"
+    chroot "${rootfs_dir}" /sbin/sfdisk "${STORAGE_DEVICE_FILE}" < "${partition_table_file}" || return 1
+
+    format_dummy_partition "/sbin/mkfs.ext4" "$(( SECTOR_SIZE * BOOT_START ))" "${STORAGE_DEVICE_FILE}" || return 1
+    format_dummy_partition "/usr/sbin/mkfs.f2fs" "$(( SECTOR_SIZE * ROOTFS_START ))" "${STORAGE_DEVICE_FILE}" || return 1
+    format_dummy_partition "/usr/sbin/mkfs.f2fs" "$(( SECTOR_SIZE * SPOOL_START ))" "${STORAGE_DEVICE_FILE}" || return 1
+
+    unlink "${partition_table_file}"
```
oliv3r: because you create this in this function, i understand why you clean it up at the end again. However if any of the previous commands fail, this will never get cleaned.

tests/toolbox_image.sh
```diff
@@ -13,6 +24,55 @@ result=0
...
+
+    if grep "${loop_device}" "/proc/mounts"; then
+        umount "${rootfs_dir}${loop_device}"
+        unlink "${rootfs_dir}${loop_device}"
+    fi
+}
+
+create_dummy_storage_device()
+{
+    partition_table_file="${rootfs_dir}/tmp/partition_table"
```
oliv3r: why store this in a file, if it is created AND removed in this function only. Just store the data in a variable and use the variable as stdin if possible. Unless it is really needed avoid storing temporary data in files

tests/toolbox_image.sh
```diff
@@ -51,6 +113,10 @@ teardown()
         return
     fi
 
+    if [ -r "${STORAGE_DEVICE_FILE}" ]; then
```
oliv3r: during cleanup, i think we can use -e or -f; the file does not have to be readable by the user. I don't even think we need to have write access to the file, we just need write access to the upper laying directory

tests/toolbox_image.sh
```diff
@@ -51,6 +113,10 @@ teardown()
...
+        unlink "${STORAGE_DEVICE_FILE}"
```
oliv3r: could it be the case that it is still mounted? if not, then this is obviously fine

tests/toolbox_image.sh
```diff
@@ -81,6 +147,9 @@ failure_exit()
     echo "  sudo sh -c '\\"
     echo "    umount '${rootfs_dir}/${ARM_EMU_BIN}' && \\"
     echo "    unlink '${rootfs_dir}/${ARM_EMU_BIN}' && \\"
+    echo "    for loop_device in /dev/loop[0-9]; do ( if grep '${loop_device}' '/proc/mounts'; \\ "
+    echo "    then sudo umount '${rootfs_dir}${loop_device}'; fi ); done && \\"
```
oliv3r: i put a sudo in your sudo so you could sudo :D

tests/toolbox_image.sh
```diff
@@ -81,6 +147,9 @@ failure_exit()
...
+    echo "    for loop_device in /dev/loop[0-9]; do unlink '${rootfs_dir}${loop_device}'; done && \\"
```
oliv3r: why two loops?
```sh
    echo "    for loop_device in /dev/loop*; do \\"
    echo "        if grep '\${loop_device}' '/proc/mounts'; then \\"
    echo "            umount '${rootfs_dir}\${loop_device}'; \\"
    echo "        fi; \\"
    echo "        if [ -f '${rootfs_dir}\${loop_device}' ]; then \\"
    echo "            unlink '${rootfs_dir}\${loop_device}' ]; \\"
    echo "        fi; \\"
    echo "  done && \\"
```
you want to escape the $ on loop_device, to prevent it from resolving, but ${rootfs_dir} should be resolved.

more importantly however:
loop[0-9] won't umount loop10
I am quite unhappy with the /dev/loop* however as it will umount ALL loop devices, INCLUDING the ones I have unrelated to this test!
I think what would be better, is to get all loop mount devices from /proc/mount that match rootfs_dir.
something like
```sh
loop_devices="$(grep '"${rootfs_dir}"*/dev/loop' /proc/mounts)"
```
Not sure if the * works without special quotes or anything, and not sure if you can do this with busybox grep. I think you need grep -E (to do regex) and then the asterix becomes .* (e.g. match any character inbetween) (if the order is even right). But that will atleast all valid devices!

see https://www.thegeekstuff.com/2011/10/grep-or-and-not-operators for reference

tests/toolbox_image.sh
```diff
@@ -159,6 +228,26 @@ test_system_update_entrypoint()
     test -x "${rootfs_dir}${SYSTEM_UPDATE_ENTRYPOINT}"
 }
 
+test_execute_resize_partition_shrink_spool_ok()
+{
+    partition_table_file="/tmp/partition_table"
```
oliv3r: why not use chroot ${rootfs_dir} mktemp? (if you can't store it in a variable instead)
remember however that you also need to clean it up when things fail!

tests/toolbox_image.sh
```diff
@@ -159,6 +228,26 @@ test_system_update_entrypoint()
...
+    chroot "${rootfs_dir}" /sbin/sfdisk -d "${STORAGE_DEVICE_FILE}" > "${rootfs_dir}${partition_table_file}"
+
+    spool_size="$(( SECTOR_TOTAL - SPOOL_START ))"
+    new_spool_size="$(( spool_size / 2 ))"
+    new_spool_start="$(( SPOOL_START + new_spool_size ))"
```
oliv3r: while completely relevant here, this is a note to ourselves, the final 'actual' partition table will need to account for partition and file-system alignment to flash erase sizes!

oliv3r (changes requested):

tests/toolbox_image.sh
raysiudak: Nope

tests/toolbox_image.sh
raysiudak: create_dummy_storage_device is called in setup and the images file is destroyed in teardown

tests/toolbox_image.sh
raysiudak: Because EOT end should start line column 1, i think it makes it more readable if the start is also on the first line.

tests/toolbox_image.sh
raysiudak: Yeah, forgot to remove this. Before moving this code to the setup, i wanted the function to return control instead of stopping the script, so that it was possible to print the failure exit text.

tests/toolbox_image.sh
raysiudak: No cannot because we have no environment when running the command.

tests/toolbox_image.sh
raysiudak: Still being mounted would be an error in the setup and then the script should stop

tests/toolbox_image.sh
raysiudak: Yes. 
The code is changed, the scenario where failrue exit is printed and the loop devices are mounted is not possible anymore, therefore removed the this code.

tests/toolbox_image.sh
raysiudak: Putting it in a variable was what i had before we decided to do it with the toolbox tools in the chroot -ed environment.    
I don't think it is possible, normally you would 'echo partition_table | fdisk image_file' i don't see how this can be done in chroot.

tests/toolbox_image.sh
raysiudak: See comment down the line, i don't see how it is possible using chroot

tests/buildenv.sh
```diff
@@ -69,7 +68,7 @@ check_emulation_support()
         return
     fi
 
-    echo "ok"
+    echo "${ARM_EMU_BIN} ok"
```
oliv3r: :+1:

tests/buildenv.sh
```diff
@@ -68,7 +68,7 @@ check_emulation_support()
         return
     fi
 
-    echo "${ARM_EMU_BIN} ok"
+    echo "${ARM_EMU_BIN}: ok"
```
oliv3r: i think this change was part of your other fixup ;)

tests/toolbox_image.sh
```diff
@@ -7,13 +7,16 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
 SYSTEM_UPDATE_ENTRYPOINT="/sbin/startup.sh"
 DISK_PREPARE_COMMAND="/sbin/prepare_disk.sh"
 
-# Test storage device parameters
-STORAGE_DEVICE_FILE="/tmp/storage_device.img"
-SECTOR_SIZE="512"
-SECTOR_TOTAL="7553024"
+# Test storage device parameters.
+# All storage start and and parameters are sector numbers. actual byte positions
```
oliv3r: and end
instead of
and and

sector numbers. Actual byte

tests/toolbox_image.sh
```diff
@@ -7,13 +7,16 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
...
+# are calculated by multiplying the sector number with the sector size.
+STORAGE_DEVICE_IMG="/tmp/storage_device.img"
+SECTOR_SIZE_B="512"
+STORAGE_END="7553024"
```
oliv3r: To be consistent with the above, call it STORAGE_DEVICE_IMG_END (and sort it next to it)
also a comment  behind it, indicating the true size would be helpful

tests/toolbox_image.sh
```diff
@@ -7,13 +7,16 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
...
 
-# Test storage device parameters
-STORAGE_DEVICE_FILE="/tmp/storage_device.img"
-SECTOR_SIZE="512"
-SECTOR_TOTAL="7553024"
+# Test storage device parameters.
+# All storage start and and parameters are sector numbers. actual byte positions
+# are calculated by multiplying the sector number with the sector size.
+STORAGE_DEVICE_IMG="/tmp/storage_device.img"
+SECTOR_SIZE_B="512"
```
oliv3r: what does the B mean here?

besides, why did you change it? I think it was fine before :)

tests/toolbox_image.sh
oliv3r: i'm not seeing the cleanup of the file however now.
Maybe pre-generate a in setup() and clean it up in teardown().

oliv3r (changes requested):

tests/toolbox_image.sh
raysiudak: BYTES but ok i will change it back

Dockerfile
```diff
@@ -4,7 +4,12 @@ LABEL Maintainer="software-embedded-platform@ultimaker.com" \
       Comment="Ultimaker update-tools filesystem"
 
 RUN apk add --no-cache \
+        e2fsprogs \
+        f2fs-tools \
+        sed \
```
oliv3r: i still think this is unnecessary for now as afaik the busybox variant does exactly what you request it to do. but it's okay to leave it in as it makes you happy :)

tests/buildenv.sh
```diff
@@ -5,8 +5,8 @@ set -eu
 ARM_EMU_BIN="${ARM_EMU_BIN:-}"
 BINFMT_MISC="${BINFMT_MISC:-/proc/sys/fs/binfmt_misc/}"
 
-FILESYSTEMS="overlay squashfs tmpfs"
-COMMANDS="apk mksquashfs xz"
+FILESYSTEMS="ext4 f2fs overlay squashfs tmpfs"
+COMMANDS="apk mkfs.ext4 mkfs.f2fs mksquashfs sed sfdisk xz"
```
oliv3r: :+1: irregardless of which variant of sed is used, it should be in this list

tests/toolbox_image.sh
```diff
@@ -3,8 +3,23 @@
...
+
 SYSTEM_UPDATE_ENTRYPOINT="/sbin/startup.sh"
+DISK_PREPARE_COMMAND="/sbin/prepare_disk.sh"
+
+# Test storage device parameters.
+# All storage start and and parameters are sector numbers. actual byte positions
+# are calculated by multiplying the sector number with the sector size.
+STORAGE_DEVICE_IMG="/tmp/storage_device.img"
+SECTOR_SIZE="512"
+STORAGE_DEVICE_END="7553024"
```
oliv3r: still missing comment at the end about the size
```sh
STORAGE_DEVICE_END="7553024" # sectors, about 3.6 GiB
```

tests/toolbox_image.sh
```diff
@@ -3,8 +3,23 @@
...
+BOOT_START="2048"
+ROOTFS_START="67584"
```
oliv3r:
```sh
ROOTFS_START="67584" # offset 33 MiB
```
or something like that

tests/toolbox_image.sh
```diff
@@ -3,8 +3,23 @@
...
+USERDATA_START="1998848"
```
oliv3r:
```sh
USERDATA_START="1998848" # offset 976 MiB
```

tests/toolbox_image.sh
```diff
@@ -13,6 +28,34 @@ result=0
...
+{
+    echo "Creating test image: '${rootfs_dir}${STORAGE_DEVICE_IMG}'."
+    dd if="/dev/zero" of="${rootfs_dir}${STORAGE_DEVICE_IMG}" bs="1" count="0" \
+        seek="$((SECTOR_SIZE * STORAGE_DEVICE_END))"
+
+    echo "writing partition table:"
+
+    sfdisk "${rootfs_dir}${STORAGE_DEVICE_IMG}" << \
+______________________________________________________________________________________
+unit: sectors
```
oliv3r: for completeness sake, i'd add the 'label: dos' marker above the unit as well, to ensure we do not generate GPT partitions (we currently cannot support proper GPT on our platform)

tests/toolbox_image.sh
```diff
@@ -40,7 +83,14 @@ setup()
...
     ln -s ../proc/self/mounts "${rootfs_dir}/etc/mtab"
 
+    mount -t devtmpfs none "/dev"
+
     dd if=/dev/zero of="${rootfs_dir}/${TMP_TEST_IMAGE_FILE}" bs=32M count=4 2> /dev/null
+
+    create_dummy_storage_device
+
+    sfdisk -d "${LOOP_STORAGE_DEVICE}" > "${rootfs_dir}${PARTITION_TABLE_FILE}"
+
```
oliv3r: one newline too much

tests/toolbox_image.sh
```diff
@@ -40,7 +83,14 @@ setup()
     mount --bind /proc "${rootfs_dir}/proc" 1> /dev/null
     ln -s ../proc/self/mounts "${rootfs_dir}/etc/mtab"
 
+    mount -t devtmpfs none "/dev"
```
oliv3r: Is this /dev device to be intended to be abailable in the chrooted environment? or in the host.

If it is supposed to be available in the chroot, then this should be mounted at "${rootfs_dir}/dev"
if it is supposed to be mounted in the host, we have to first check if it is not already mounted, (which is only unlikely if we are in the docker environment)

tests/toolbox_image.sh
```diff
@@ -60,6 +122,10 @@ teardown()
         fi
     fi
 
+    if grep -q "/dev" /proc/mounts; then
+        umount "/dev" || result=1
```
oliv3r: same here, if it is the chrooted env. then rootfs_dir is missing, otherwise ... tricky how to solve this, as we do not want to umount our hosts /dev. But since the command below does /proc for rootfs_dir, my guesstimate says its missing rootfs_dir?

tests/toolbox_image.sh
```diff
@@ -81,6 +147,9 @@ failure_exit()
     echo "  sudo sh -c '\\"
     echo "    umount '${rootfs_dir}/${ARM_EMU_BIN}' && \\"
     echo "    unlink '${rootfs_dir}/${ARM_EMU_BIN}' && \\"
+    echo "    losetup -d '${LOOP_STORAGE_DEVICE}' && \\"
+    echo "    unlink '${STORAGE_DEVICE_IMG}' && \\"
+    echo "    umount '/dev' && \\"
```
oliv3r: as above missing rootfs_dir?

tests/toolbox_image.sh
```diff
@@ -107,7 +178,7 @@ run_test()
             failure_exit
         fi
     fi
-    printf '\n'
+    echo "______________________________________________________________________________________"
```
oliv3r: :+1:

tests/toolbox_image.sh
```diff
@@ -97,6 +166,8 @@ run_test()
 {
     setup
 
+    echo "______________________________________________________________________________________"
+    echo
```
oliv3r: :trophy:  ah, that's even nicer then echo ""

tests/toolbox_image.sh
```diff
@@ -159,6 +230,25 @@ test_system_update_entrypoint()
     test -x "${rootfs_dir}${SYSTEM_UPDATE_ENTRYPOINT}"
 }
 
+test_execute_resize_partition_grow_rootfs_ok()
+{
+    userdata_size="$((STORAGE_DEVICE_END - USERDATA_START))"
+    new_userdata_size="$((userdata_size / 2))"
```
oliv3r: for this test case it won't matter, but always be weary of divisions as the the result will be truncated (integer div only)

tests/toolbox_image.sh
```diff
@@ -159,6 +230,25 @@ test_system_update_entrypoint()
     test -x "${rootfs_dir}${SYSTEM_UPDATE_ENTRYPOINT}"
 }
 
+test_execute_resize_partition_grow_rootfs_ok()
+{
+    userdata_size="$((STORAGE_DEVICE_END - USERDATA_START))"
```
oliv3r: don't we need $ markers for the variables? I guess not, i think shell check even says we don't need them! I probably have to remove them in my other script then ;)

tests/toolbox_image.sh
```diff
@@ -159,6 +230,25 @@ test_system_update_entrypoint()
...
+    new_userdata_size="$((userdata_size / 2))"
+    new_userdata_start="$((USERDATA_START + new_userdata_size))"
+
+    sed -i "s/p3.*type/p3 : start=     ${new_userdata_start}, size=     ${userdata_size}, type/" \
```
oliv3r: can you add a simple oneliner comment above BRIEFLY explaining the regular expression?
(I think its something like replace partition something?)

tests/toolbox_image.sh
```diff
@@ -159,6 +230,25 @@ test_system_update_entrypoint()
...
+        "${rootfs_dir}${PARTITION_TABLE_FILE}"
+
+    sha512sum "${rootfs_dir}${PARTITION_TABLE_FILE}" > "${rootfs_dir}${PARTITION_TABLE_FILE}.sha512" || return 1
```
oliv3r: i kind of started to agree with your use of return 1 as documentation. Maybe if we ever get time to do EMP-374 this will be more apparant as well

oliv3r (changes requested):

tests/toolbox_image.sh
raysiudak: It is intended to be on the host. Ok will and a check

tests/toolbox_image.sh
raysiudak: Try to shellcheck "$((${STORAGE_DEVICE_END} - ${USERDATA_START}))"

tests/toolbox_image.sh
```diff
@@ -83,14 +84,15 @@ setup()
     mount --bind /proc "${rootfs_dir}/proc" 1> /dev/null
     ln -s ../proc/self/mounts "${rootfs_dir}/etc/mtab"
 
-    mount -t devtmpfs none "/dev"
+    if ! grep -q "/dev" /proc/mounts; then
+        mount -t devtmpfs none "/dev"
```
oliv3r: can we look at this together? I still think it should be ${rootfs_dir}/dev

tests/toolbox_image.sh
```diff
@@ -232,10 +234,11 @@ test_system_update_entrypoint()
 
 test_execute_resize_partition_grow_rootfs_ok()
 {
-    userdata_size="$((STORAGE_DEVICE_END - USERDATA_START))"
+    userdata_size="$((STORAGE_DEVICE_SIZE - USERDATA_START))"
     new_userdata_size="$((userdata_size / 2))"
     new_userdata_start="$((USERDATA_START + new_userdata_size))"
 
+    # In every line look for a string "p3<don't care>type" and replace it.
```
oliv3r: and replace it to ... why? I'm guessing to match the end of the disk? If that is the case, I think there's a special marker you can put in the file 'start until end'. for the CLI interface, you can do sfdisk start,,L iirc (Granted my script won't work with that yet :) ) if that was the intention. i'm fine with it being explicit as well, you decide.

test/toolbox_image.sh
```diff
@@ -81,6 +158,9 @@ failure_exit()
     echo "  sudo sh -c '\\"
     echo "    umount '${rootfs_dir}/${ARM_EMU_BIN}' && \\"
     echo "    unlink '${rootfs_dir}/${ARM_EMU_BIN}' && \\"
+    echo "    losetup -d '${LOOP_STORAGE_DEVICE}' && \\"
+    echo "    unlink '${STORAGE_DEVICE_IMG}' && \\"
+    echo "    umount '/dev' && \\"
```
oliv3r: you forgot to add your umount steps here :)
```sh
echo "    umount '${rootfs_dir}/dev'"
if "${is_dev_setup_mounted}"; then
    echo "    umount "/dev"
fi
```

oliv3r (changes requested):

oliv3r (approved):

PR #19, opened 2018-08-21
WIP EMP-389 prepare disk tests

raysiudak:


build_for_ultimaker.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
-
```
oliv3r: excellent idea! but could you leave the whitespace (below the copyright text) for the vertically challenged ;

test/buildenv.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
-
```
oliv3r: ditto

test/toolbox_image.sh
```diff
@@ -51,9 +57,9 @@ ________________________________________________________________________________
 
     LOOP_STORAGE_DEVICE="$(losetup --show --find --partscan "${rootfs_dir}${STORAGE_DEVICE_IMG}")"
 
-    mkfs.ext4 -q "${LOOP_STORAGE_DEVICE}p1"
-    mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p2"
-    mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p3"
+    mkfs.ext4 -q "${LOOP_STORAGE_DEVICE}p1" && fsck -y "${LOOP_STORAGE_DEVICE}p1"
+    mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p2" && fsck "${LOOP_STORAGE_DEVICE}p2"
+    mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p3" && fsck "${LOOP_STORAGE_DEVICE}p3"
```
oliv3r: i take it the f2fs tool doesn't have a -y argument? I'd also add the --force and the 'dry-run' option to ensure fsck is actually run

test/toolbox_image.sh
```diff
@@ -183,9 +189,12 @@ run_test()
     echo "______________________________________________________________________________________"
     echo
     echo "Run: ${1}"
+    echo
     if "${1}"; then
+        echo
         echo "Result - OK"
     else
+        echo
```
oliv3r: de-duplicate these echo's, if you have the exact same code in both; maybe it's better to put it above. (Do we really need 2 blank lines before the result? (I'm ok with 2 blank lines)

test/toolbox_image.sh
oliv3r: why not call the test integrity helper after formatting? isn't that what the helper is for? your duplicating its functionality here ;)

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
     test -x "${rootfs_dir}${SYSTEM_UPDATE_ENTRYPOINT}"
 }
 
-test_execute_resize_partition_grow_rootfs_ok()
+helper_test_disk_integrity()
```
oliv3r: i'd almost argue this is not a helper, but just a regular function!

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
...
 
+helper_execute_prepare_disk()
+{
     # sha512sum is executed in the chroot environment to avoid that rootfs dir is prefixed to the file path.
     chroot "${rootfs_dir}" sha512sum "${PARTITION_TABLE_FILE}" > "${rootfs_dir}${PARTITION_TABLE_FILE}.sha512" || return 1
     chroot "${rootfs_dir}" "${DISK_PREPARE_COMMAND}" -t "${PARTITION_TABLE_FILE}" "${LOOP_STORAGE_DEVICE}" || return 1
 
     verification_partition_table_file="/tmp/verification_partition_table"
-    sfdisk -d "${LOOP_STORAGE_DEVICE}" > "${rootfs_dir}${verification_partition_table_file}"
+    sfdisk -d "${LOOP_STORAGE_DEVICE}" > "${rootfs_dir}${verification_partition_table_file}" || return 1
```
oliv3r: add the || return to the caller, or is that not possible due to the if [ "${test}" ]; because if that's the case, we really have to change the way we do that bit

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
...
 }
 
+test_execute_disk_prepare_grow_boot_overlapping_rootfs_nok()
+{
+    new_boot_size="$((ROOTFS_START - BOOT_START + 1))"
+
+    sed -i "s/p2.*type/p1 : start=     ${BOOT_START}, size=     ${new_boot_size}, type/" \
+        "${rootfs_dir}${PARTITION_TABLE_FILE}"
+
+    helper_execute_prepare_disk || return 0
```
oliv3r: seeing it here makes it even uglier imo

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
...
+
+    # In every line in the partition table look for a string "p2<don't care>type" and replace it with
+    # with new the new disk partition parameters defined above.
+    sed -i "s/p1.*type/p1 : start=     ${BOOT_START}, size=      ${new_boot_size}, type/" \
+        "${rootfs_dir}${PARTITION_TABLE_FILE}"
+    sed -i "s/p2.*type/p2 : start=     ${new_rootfs_start}, size=      ${new_rootfs_size}, type/" \
+        "${rootfs_dir}${PARTITION_TABLE_FILE}"
+
+    helper_execute_prepare_disk || return 1
+    helper_test_disk_integrity || return 1
```
oliv3r: and now i want to strongly argue that they are not really 'helpers ' in that sense, but just regular functions we are calling

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
...
     sed -i "s/label-id:.*//" "${rootfs_dir}${PARTITION_TABLE_FILE}"
     sed -i "s/label-id:.*//" "${rootfs_dir}${verification_partition_table_file}"
     diff "${rootfs_dir}${PARTITION_TABLE_FILE}" "${rootfs_dir}${verification_partition_table_file}" || return 1
 }
 
+test_execute_disk_prepare_grow_boot_overlapping_rootfs_nok()
+{
+    new_boot_size="$((ROOTFS_START - BOOT_START + 1))"
+
+    sed -i "s/p2.*type/p1 : start=     ${BOOT_START}, size=     ${new_boot_size}, type/" \
```
oliv3r: document the regular expression so that we know what's happening in 2 years down the line
```sh
# Replace p2, the root partition with a larger p1, the boot partition
```
or something is enough and clear?

While this regex works quite well, i think a more complex one could have also helped by ONLY replacing the 'size' parameter value, rather then the entire line; but that's beyond my regex skills :p things like https://www.unix.com/shell-programming-and-scripting/172036-replace-specific-field-specific-line-sed-awk.html do have answer/solution, but they are not fun to read :p

I really think we need this because just glancing at it, I'm having trouble deciphering it. replace p2 with p1? huh?

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
...
 
     # Remove the identifiers in the header because they will always change.
     sed -i "s/label-id:.*//" "${rootfs_dir}${PARTITION_TABLE_FILE}"
     sed -i "s/label-id:.*//" "${rootfs_dir}${verification_partition_table_file}"
     diff "${rootfs_dir}${PARTITION_TABLE_FILE}" "${rootfs_dir}${verification_partition_table_file}" || return 1
 }
 
+test_execute_disk_prepare_grow_boot_overlapping_rootfs_nok()
+{
+    new_boot_size="$((ROOTFS_START - BOOT_START + 1))"
```
oliv3r: While + 1 is okay-ish, why not do $((BOOT_SIZE + (${RANDOM} % ROOTFS_END) + 1)

to be fair, a proper test would then run this function RANDOM times to truly test it a few times.

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
...
+        "${rootfs_dir}${PARTITION_TABLE_FILE}"
+
+    helper_execute_prepare_disk || return 0
+}
+
+test_execute_disk_prepare_grow_rootfs_overlapping_userdata_nok()
+{
+    new_rootfs_size="$((USERDATA_START - ROOTFS_START + 1))"
+
+    sed -i "s/p2.*type/p2 : start=     ${ROOTFS_START}, size=     ${new_rootfs_size}, type/" \
```
oliv3r: hmm, p2 again? was the previous one a typo?

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
...
+        "${rootfs_dir}${PARTITION_TABLE_FILE}"
+
+    helper_execute_prepare_disk || return 0
+}
+
+test_execute_disk_prepare_grow_boot_invalid_start_nok()
+{
+    new_boot_start="2047"
```
oliv3r: same here, 2048 is what we want, anything below 2048 is wrong, lets randomize it for broader test coverage.

If we feel crazy, we could even call this function 0 - 2047 times, but something more random is a bit safer

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
...
+    sed -i "s/p3.*type/p3 : start=     ${new_userdata_start}, size=     ${new_userdata_size}, type/" \
+        "${rootfs_dir}${PARTITION_TABLE_FILE}"
+
+    helper_execute_prepare_disk || return 1
+    helper_test_disk_integrity || return 1
+}
+
+test_execute_disk_prepare_grow_boot_ok()
+{
+    rootfs_size="$((USERDATA_START - ROOTFS_START))"
```
oliv3r: maybe use this pattern on the other functions as well?

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
...
+    new_rootfs_size="$((rootfs_size / 2))"
```
oliv3r: while this is a good test, don't we want to maybe grow the rootfs as well? IN any case the function name doens't fully map to this. You only want to grow the boot. Granted to be able to do that somethings got to give.

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
...
+    helper_execute_prepare_disk || return 1
+    helper_test_disk_integrity || return 1
+  }
+
+test_execute_disk_prepare_sha512_nok()
+{
+    partition_table_file="/tmp/partition_table"
+    sfdisk -d "${rootfs_dir}${STORAGE_DEVICE_IMG}" > "${rootfs_dir}${partition_table_file}"
+    chroot "${rootfs_dir}" sha512sum "${PARTITION_TABLE_FILE}" > "${rootfs_dir}${PARTITION_TABLE_FILE}.sha512"
+    echo "some extra data" >> "${rootfs_dir}${partition_table_file}"
```
oliv3r: note, that this test will fail because the 'extra data' will not validate, not because the shasum is invalid.

so i'd say, rewrite ($RANDOM) character(s) in the sha15sum?

test/toolbox_image.sh
```diff
@@ -254,36 +263,124 @@ test_system_update_entrypoint()
...
+
+    helper_execute_prepare_disk || return 1
+    helper_test_disk_integrity || return 1
+  }
+
+test_execute_disk_prepare_sha512_nok()
+{
+    partition_table_file="/tmp/partition_table"
+    sfdisk -d "${rootfs_dir}${STORAGE_DEVICE_IMG}" > "${rootfs_dir}${partition_table_file}"
+    chroot "${rootfs_dir}" sha512sum "${PARTITION_TABLE_FILE}" > "${rootfs_dir}${PARTITION_TABLE_FILE}.sha512"
```
oliv3r:
```sh
# Workaround to avoid having to deal with different workspaces, run sha512sum from within the ${rootfs_dir} workspace.
```

test/toolbox_image.sh
```diff
@@ -341,7 +438,13 @@ run_test test_execute_resizef2fs
 run_test test_execute_mount
 run_test test_execute_rsync
 run_test test_system_update_entrypoint
+run_test test_execute_disk_prepare_grow_boot_overlapping_rootfs_nok
+run_test test_execute_disk_prepare_grow_rootfs_overlapping_userdata_nok
+run_test test_execute_disk_prepare_grow_boot_invalid_start_nok
+run_test test_execute_disk_prepare_grow_beyond_disk_end_nok
 run_test test_execute_resize_partition_grow_rootfs_ok
```
oliv3r: change on naming convention?

oliv3r (changes requested):

test/toolbox_image.sh
raysiudak: Yeah they have grown to become a 'real' functions, i do agree.

test/toolbox_image.sh
raysiudak: Yes i will fix these in the next commit, because the function is introduced in that commit.

raysiudak: Yes i will fix these in the next commit, because the function is introduced in that commit.

test/toolbox_image.sh
raysiudak: Ok it is a bit confusing. What i want in this function is that the result of executing prepare_disk is returned when it fails and if that the result is checked when it does not fail. Therefore i removed the '|| return' from the other test setup statements, because this is not right. The actual tests will fail if the test setup has errors.

test/toolbox_image.sh
raysiudak: Yeah it was a typo

test/toolbox_image.sh
raysiudak: Yes

test/toolbox_image.sh
raysiudak: Added a random value below 2048

test/toolbox_image.sh
raysiudak: The test function says "test execution of disk_prepare by growing boot, expected result ok. The input is a partition table that should be correct, therefore the function modifies the boot "and" the rootfs partitions.

test/toolbox_image.sh
raysiudak: No im am making the partition file corrupt so that the md512sum does not validate anymore.

test/toolbox_image.sh
```diff
@@ -57,11 +66,17 @@ ________________________________________________________________________________
...
+    mkfs.ext4 -q "${LOOP_STORAGE_DEVICE}p1"
+    mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p2"
+    mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p3"
 
-    echo "Successfully created dummy storage device: '${LOOP_STORAGE_DEVICE}'."
+    if test_disk_integrity; then
+        echo "Successfully created dummy storage device: '${LOOP_STORAGE_DEVICE}'."
+        return 0
+    fi
+
```
oliv3r: i'd prefer it if you'd swap these two around.

Test the integrity, and exit on failure (fail early). If everything is okay, then return normally (without error).

test/toolbox_image.sh
```diff
@@ -36,6 +36,11 @@ is_dev_setup_mounted=false
 exit_on_failure=false
 
 
+random_uint32()
+{
+    echo "$(od -vAn -N4 -tu4 < /dev/urandom)"
```
oliv3r: while doing this briefly in a terminal as:
```sh
while true; do echo "$(od -vAn -N4 -tu4 < /dev/urandom)"; done
```
I would get many random numbers, but always very high. For example:
`
 3295899462
  849972328
  761581844
   55945594
`
So I'm questioning how 'random' this is. Will it ever return  < 10? I'm guessing that what happens is, we read 1 random byte after the other and then concatenate them? I thinking they basically are not uniformly distrbuted.

oliv3r: [but busybox contains ](https://unix.stackexchange.com/a/140751) ...[shuf](https://linux.die.net/man/1/shuf)?!
```sh
shuf -i 0-"${MAX_RANGE}" -n 1
```

test/toolbox_image.sh
```diff
@@ -295,35 +300,45 @@ helper_execute_prepare_disk()
 
 test_execute_disk_prepare_grow_boot_overlapping_rootfs_nok()
 {
-    new_boot_size="$((ROOTFS_START - BOOT_START + 1))"
+    # Get a size between the current size and current + rootfs size
+    rootfs_size="$((USERDATA_START - ROOTFS_START))"
+    random_size="$(($(random_uint32) % rootfs_size))"
+    new_boot_size="$((BOOT_SIZE + random_size + 1))"
```
oliv3r: I  made a new friend today, his name is shuf! :D

test/toolbox_image.sh
```diff
@@ -295,35 +300,45 @@ helper_execute_prepare_disk()
...
-    helper_execute_prepare_disk || return 0
+    execute_prepare_disk || return 0
 }
 
 test_execute_disk_prepare_grow_rootfs_overlapping_userdata_nok()
 {
-    new_rootfs_size="$((USERDATA_START - ROOTFS_START + 1))"
+    # Get a size between the current size and current + userdata size
+    userdata_size="$((STORAGE_DEVICE_SIZE - USERDATA_START))"
+    random_size="$(($(random_uint32) % userdata_size))"
```
oliv3r: just shuf it! :p

test/toolbox_image.sh
```diff
@@ -295,35 +300,45 @@ helper_execute_prepare_disk()
...
         "${rootfs_dir}${PARTITION_TABLE_FILE}"
 
-    helper_execute_prepare_disk || return 0
+    execute_prepare_disk || return 0
 }
 
 test_execute_disk_prepare_grow_boot_invalid_start_nok()
 {
-    new_boot_start="2047"
+    new_boot_start="$(($(random_uint32) % $((BOOT_START -1))))"
```
oliv3r: missing space around the arithmetic operator - (unless you intended to say the number minus one?)

test/buildenv.sh
```diff
@@ -11,7 +11,7 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
 BINFMT_MISC="${BINFMT_MISC:-/proc/sys/fs/binfmt_misc/}"
 
 FILESYSTEMS="ext4 f2fs overlay squashfs tmpfs"
-COMMANDS="apk mkfs.ext4 mkfs.f2fs mksquashfs sed sfdisk xz"
+COMMANDS="apk mkfs.ext4 mkfs.f2fs mksquashfs od sed sfdisk xz"
```
oliv3r: make shuf your new best friend too!

test/buildenv.sh
```diff
@@ -96,10 +96,22 @@ check_command_installation()
...
+{
+    printf "urandom device file: "
+    if [ ! -c "/dev/urandom" ]; then
+        result=1
+        echo "doesn't exists"
+        return
+    fi
+    echo "exists"
+}
+
```
oliv3r, line 133: shuf don't care!

test/toolbox_image.sh
```diff
@@ -380,9 +380,9 @@ test_execute_disk_prepare_grow_boot_ok()
 
     # In every line in the partition table look for a string "p1<don't care>type" and replace it with
     # with new the new disk partition parameters defined above.
-    sed -i "s/p1.*type/p1 : start=     ${BOOT_START}, size=      ${new_boot_size}, type/" \
+    sed -i "s|${LOOP_STORAGE_DEVICE}p1.*type|${LOOP_STORAGE_DEVICE}p1 : start=        ${BOOT_START}, size=     ${new_boot_size}, type|" \
```
oliv3r: i assume the space count change before BOOT_START was intended?

oliv3r (changes requested):

test/toolbox_image.sh
oliv3r: i'm okay to keep as is, but the input partition could have the rootfs partition already reallocated to _only_ test the growing of boot of course. But as discussed, it's not worth the fixture overhead.

test/toolbox_image.sh
oliv3r: could you add this comment above sha512sum still? to document why we are running it from within the chroot

build_for_ultimaker.sh
```diff
@@ -1,5 +1,10 @@
...
+#
+# Copyright (C) 2018 Ultimaker B.V.
+# Copyright (C) 2018 Olliver Schinagl <oliver@schinagl.nl>
+# Copyright (C) 2018 Raymond Siudak <raysiudak@gmail.com>
+#
+# SPDX-License-Identifier: AGPL-3.0+
```
oliv3r, line 7: its easier on the eyes to have 1 blank line between the end of the comment block and the first line of code

test/buildenv.sh
```diff
@@ -1,12 +1,17 @@
 #!/bin/sh
-
+#
+# Copyright (C) 2018 Ultimaker B.V.
+# Copyright (C) 2018 Raymond Siudak <raysiudak@gmail.com>
+# Copyright (C) 2018 Olliver Schinagl <oliver@schinagl.nl>
+#
+# SPDX-License-Identifier: AGPL-3.0+
```
oliv3r, line 7: ditto

oliv3r (changes requested):

test/toolbox_image.sh
raysiudak: No we don't need it, but i think the output is nicer this way.

raysiudak: We don't need this, but i think the output is nicer this way

scripts/prepare_disk.sh
```diff
@@ -1,10 +1,9 @@
 #!/bin/sh
-#
-# SPDX-License-Identifier: AGPL-3.0+
-#
+
```
oliv3r: i can agree with you moving the license identifier below the copyright notices, that's good :+1: 
but keep it as one comment block:
```sh
#shebang
#
#stuff
#
# license :)
```

test/buildenv.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/sh
```
oliv3r: ditto

test/toolbox_image.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+# Copyright (C) 2018 Ultimaker B.V.
```
oliv3r: ditto

build.sh
```diff
@@ -1,4 +1,10 @@
 #!/bin/sh
+#
+# Copyright (C) 2018 Ultimaker B.V.
+# Copyright (C) 2018 Raymond Siudak <raysiudak@gmail.com>
+# Copyright (C) 2018 Olliver Schinagl <oliver@schinagl.nl>
+#
+# SPDX-License-Identifier: AGPL-3.0+
```
oliv3r, line 8: this block is the golden example :)

replicated it to stay concistent :)

test/toolbox_image.sh
```diff
@@ -29,6 +35,20 @@ result=0
 is_dev_setup_mounted=false
 exit_on_failure=false
 
+
+random_int()
+{
+    shuf -n 1 -i 0-"${1}"
+}
```
oliv3r: :100:

test/toolbox_image.sh
```diff
@@ -55,7 +75,13 @@ ________________________________________________________________________________
     mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p2"
     mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p3"
 
+    if ! test_disk_integrity; then
+        echo "Something went wrong creating dummy storage device: '${LOOP_STORAGE_DEVICE}'."
```
oliv3r: Unknown error 50; something went wrong, Abort, Retry, Fail :p

test/toolbox_image.sh
oliv3r: i'm okay with having the two new lines,

but don't put them both in the if/else, just put 2 echo's above the if; is what i meant:
```sh
echo
if "true"; then
   echo
   echo "result"
else
  echo
  echo "!result"
fi
```
would become
```sh
echo
echo
if "${1}"; then
   echo "Result - OK"
else
  echo "Result - error"
```
(or echo; echo; if you prefer it on a single line :p)

test/toolbox_image.sh
```diff
@@ -278,12 +298,112 @@ test_execute_resize_partition_grow_rootfs_ok()
...
+}
+
 usage()
 {
 cat <<-EOT
 	Usage:   "${0}" [OPTIONS] <file.img>
-	    -e   Stop consequtive tests on failure without cleanup
-	    -h   Print usage
+	    -e   Stop consecutive tests on failure without cleanup.
+	    -h   Print usage.
```
oliv3r: lol i just removed one period to stay consistent with 'ls --help' for example, where the options are not sentences.
So what will it be, are these instructive lines (like with ls --help, docker --help etc) or ... (i can't an example)?

you choose; i'll accept both solutions :)

oliv3r (changes requested):

test/toolbox_image.sh
raysiudak: I choose this

raysiudak:
This pull request has been reworked and split up into multiple other pull requests

PR #20, opened 2018-08-24
EMP-378 small buildenv/toolbox fixes

oliv3r:
A few small improvements on the toobox en buildenv test scripts. Most importantly if there is a failure in setup, teardown is called when the -e flag is not used.

### DIFF
--- a/scripts/prepare_disk.sh
+++ b/scripts/prepare_disk.sh
@@ -49,7 +49,7 @@ partition_resize()
 
         partition_end="$((start + size))"
         # sfdisk returns size in blocks, * (1024 / 512) converts to sectors
-        target_disk_end="$(($(sfdisk -s "${TARGET_DISK}" 2> /dev/null) * 2))"
+        target_disk_end="$(($(sfdisk --quiet --show-size "${TARGET_DISK}" 2> /dev/null) * 2))"
         if [ "${partition_end}" -gt "${target_disk_end}" ]; then
             echo "Partition '${label}' is beyond the size of the disk (${partition_end} > ${target_disk_end}), cannot continue."
             exit 1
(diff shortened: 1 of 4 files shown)


test/buildenv.sh
```diff
@@ -6,13 +6,29 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
...
 result=0
 
 
+check_loopdevice_support()
+{
+    printf "loopdevice support: "
+
+    PATH="${PATH}:/sbin:/usr/sbin:/usr/local/sbin" command losetup -f 1> /dev/null || { result=1; return; }
+
+    for loopdevice in /dev/loop*; do
```
raysiudak: Add quotation marks "/dev/loop"*

scripts/prepare_disk.sh
```diff
@@ -49,7 +49,7 @@ partition_resize()
 
         partition_end="$((start + size))"
         # sfdisk returns size in blocks, * (1024 / 512) converts to sectors
-        target_disk_end="$(($(sfdisk -s "${TARGET_DISK}" 2> /dev/null) * 2))"
+        target_disk_end="$(($(sfdisk --quiet -s "${TARGET_DISK}" 2> /dev/null) * 2))"
```
raysiudak: Be consistent in using single letter options or full name options

scripts/prepare_disk.sh
```diff
@@ -62,6 +62,7 @@ partition_resize()
     fi
 
     sfdisk --quiet "${TARGET_DISK}" < "${PARTITION_TABLE_FILE}"
+    sleep 1
```
raysiudak: I really do not like this because the test script will take 20+ seconds longer now, although i understand it. 
Maybe 'if there is time' we could just stay in a loop if the storage device is out of sync.

raysiudak (changes requested):

scripts/prepare_disk.sh
oliv3r: but not both options are always available! but for sfdisk they are ;)

i'll use the long options to keep the code more readable. note that this will break busybox sfdisk (once that exists)

raysiudak (approved):

PR #21, opened 2018-08-24
EMP-305 is resize needed

raysiudak:
This PR needs EMP-302_triggered_improvements merged first as it is based upon that. After that this needs to be rebased on the master to get the merge commit.

### DIFF
--- a/scripts/prepare_disk.sh
+++ b/scripts/prepare_disk.sh
@@ -32,6 +32,40 @@ is_comment()
     test -z "${1%%#*}"
 }
 
+# Returns 0 when resize is needed and 1 if not needed.
+is_resize_needed()
+{
+    current_partition_table_file="$(mktemp)"
+    sfdisk -d "${TARGET_DISK}" > "${current_partition_table_file}"
+
+    while IFS="${IFS}:=," read -r table_label _ table_start _ table_size _; do
+        if is_comment "${table_label}" || ! is_integer "${table_start}" || \
+            ! is_integer "${table_size}"; then
+            continue
+        fi
+
+        while IFS="${IFS}:=," read -r disk_label _ disk_start _ disk_size _; do
+            if is_comment "${disk_label}" || ! is_integer "${disk_start}" || \
+                ! is_integer "${disk_size}"; then
+                continue
+            fi
+
+            if [ "${table_label}" != "${disk_label}" ]; then
+                continue
+            fi
+
+            if [ "${table_start}" -ne "${disk_start}" ] || \
+               [ "${table_size}" -ne "${disk_size}" ]; then
+                unlink "${current_partition_table_file}"
+                return 0
+            fi
+        done < "${current_partition_table_file}"
+    done < "${PARTITION_TABLE_FILE}"
+
+    unlink "${current_partition_table_file}"
+    return 1
+}
+
 partition_sync()
 {
     i=10
@@ -200,6 +234,12 @@ fi
 if [ ! -b "${TARGET_DISK}" ]; then
     echo "Error, block device '${TARGET_DISK}' does not exist."
     exit 1
+fi
+
+resize_needed="$(is_resize_needed; echo "${?}")"
+if [ "${resize_needed}" -eq 1 ]; then
+    echo "Partition resize not required."
+    exit 0
 fi
 
 partition_resize
--- a/scripts/prepare_disk.sh
+++ b/scripts/prepare_disk.sh
@@ -37,6 +37,7 @@ partition_sync()
     done
 
     echo "Partprobe failed, giving up."
+
     return 1
 }
 
@@ -193,6 +194,11 @@ TARGET_DISK="${*}"
 
 if [ ! -r "${PARTITION_TABLE_FILE}" ]; then
     echo "Unable to read partition table file '${PARTITION_TABLE_FILE}', cannot continue."
+    exit 1
+fi
+
+if [ ! -b "${TARGET_DISK}" ]; then
+    echo "Error, block device '${TARGET_DISK}' does not exist."
     exit 1
 fi
 
--- a/scripts/prepare_disk.sh
+++ b/scripts/prepare_disk.sh
@@ -22,6 +22,16 @@ usage()
     echo "NOTE: This script is destructive and will destroy your data."
 }
 
+is_integer()
+{
+    test "${1}" -eq "${1}" 2> /dev/null
+}
+
+is_comment()
+{
+    test -z "${1%%#*}"
+}
+
 partition_sync()
 {
     i=10
@@ -110,16 +120,6 @@ partitions_format()
 
 partition_resize()
 {
-    is_integer()
-    {
-        test "${1}" -eq "${1}" 2> /dev/null
-    }
-
-    is_comment()
-    {
-        test -z "${1%%#*}"
-    }
-
     if ! sha512sum -csw "${PARTITION_TABLE_FILE}.sha512"; then
         echo "Error processing partition table: crc error."
         exit 1
--- a/test/toolbox_image.sh
+++ b/test/toolbox_image.sh
@@ -336,6 +336,12 @@ test_execute_resize_partition_grow_rootfs_ok()
     test_disk_integrity || return 1
 }
 
+test_execute_disk_prepare_resize_not_needed_ok()
+{
+    execute_prepare_disk || return 1
+    test_disk_integrity || return 1
+}
+
 usage()
 {
 cat <<-EOT
@@ -403,6 +409,8 @@ run_test test_execute_rsync
 run_test test_system_update_entrypoint
 run_test test_execute_disk_prepare_sha512_nok
 run_test test_execute_resize_partition_grow_rootfs_ok
+run_test test_execute_disk_prepare_resize_not_needed_ok
+
 
 if [ "${result}" -ne 0 ]; then
    echo "ERROR: There where failures testing '${ROOTFS_IMG}'."
--- a/build.sh
+++ b/build.sh
@@ -17,7 +17,7 @@ ALPINE_REPO="${ALPINE_REPO:-http://dl-cdn.alpinelinux.org/alpine}"
 TOOLBOX_IMAGE="${TOOLBOX_IMAGE:-um-update_toolbox.xz.img}"
 ROOTFS_DIR="${BUILD_DIR}/rootfs"
 
-COMMANDS="blkid busybox e2fsprogs-extra f2fs-tools rsync sfdisk"
+PACKAGES="blkid busybox e2fsprogs-extra f2fs-tools rsync sfdisk"
 
 cleanup()
 {
@@ -93,7 +93,7 @@ bootstrap_rootfs()
     # allow word splitting for ${PACKAGES}
     apk --root "${ROOTFS_DIR}" --update-cache \
        add --allow-untrusted --initdb --arch "${ARCH}" \
-       ${COMMANDS}
+       ${PACKAGES}
 
     # Add baselayout etc files
     echo "Adding baselayout"


build_for_ultimaker.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
-
+#
+# Copyright (C) 2018 Ultimaker B.V.
+# Copyright (C) 2018 Olliver Schinagl <oliver@schinagl.nl>
+# Copyright (C) 2018 Raymond Siudak <raysiudak@gmail.com>
+#
+# SPDX-License-Identifier: AGPL-3.0+
 set -eu
```
oliv3r: missing whitespace

scripts/prepare_disk.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-#
+
```
oliv3r: :-1:

scripts/prepare_disk.sh
```diff
@@ -22,6 +22,52 @@ usage()
...
 }
 
+is_resize_needed()
+{
+    is_comment()
+    {
+        test -z "${1%%#*}"
+    }
+
+    # Check if value is an integer, by using an inverse integer comparison test.
```
oliv3r: actually, it's not inverse anymore. before we had to test if it was NOT an integer to be able to skip the line, now we actually test if it IS an integer. So we now just do an integer comparision test.
anyway, with the helper function, i'd argue ti's quite self explanatory? 1 = 1 :)

scripts/prepare_disk.sh
```diff
@@ -22,6 +22,52 @@ usage()
...
+    is_integer()
+    {
+        test "${1}" -eq "${1}" 2> /dev/null
+    }
+
+    current_partition_table_file="$(mktemp)"
```
oliv3r: remember to remove this file during cleanup!

scripts/prepare_disk.sh
```diff
@@ -22,6 +22,52 @@ usage()
...
+    sfdisk -d "${TARGET_DISK}" > "${current_partition_table_file}"
+
+    exit_loop=false
+    while IFS="${IFS}:=," read -r label _ start _ size _; do
+        if is_comment "${label}" || ! is_integer "${start}"; then
+            continue
+        fi
+
+        while IFS="${IFS}:=," read -r label2 _ start2 _ size2 _; do
+            if is_comment "${label2}" || ! is_integer "${start2}"; then
```
oliv3r: label and label2, maybe also use disk_label and table_label?

scripts/prepare_disk.sh
```diff
@@ -22,6 +22,52 @@ usage()
...
+    {
+        test "${1}" -eq "${1}" 2> /dev/null
+    }
+
+    current_partition_table_file="$(mktemp)"
+    sfdisk -d "${TARGET_DISK}" > "${current_partition_table_file}"
+
+    exit_loop=false
+    while IFS="${IFS}:=," read -r label _ start _ size _; do
+        if is_comment "${label}" || ! is_integer "${start}"; then
```
oliv3r: i'd also abort if the size is not an integer?

scripts/prepare_disk.sh
```diff
@@ -22,6 +22,52 @@ usage()
...
+            fi
+
+            if [ "${label}" != "${label2}" ]; then
+                continue
+            fi
+
+            if [ "${start}" -ne "${start2}" ] || \
+               [ "${size}" -ne "${size2}" ]; then
+                exit_loop=true
+                return
```
oliv3r: i assume break cannot be used as break will only exit one loop? and return will somehow exit both loops? Maybe add a short comment about this behavior? `# break both while loops`

scripts/prepare_disk.sh
```diff
@@ -22,6 +22,52 @@ usage()
...
+            fi
+        done < "${current_partition_table_file}"
+    done < "${PARTITION_TABLE_FILE}"
+
+    if "${exit_loop}"; then
```
oliv3r: the name of the variable is not 'accurate' anymore, earlier you used it as a condition to exit the loop, now you actually use it as a changes_needed/no_changes_needed flag?

as such, I'd recommend `if ! "${changes_needed"; then return 0; fi`

scripts/prepare_disk.sh
```diff
@@ -22,6 +22,52 @@ usage()
...
+            if is_comment "${label2}" || ! is_integer "${start2}"; then
+                continue
+            fi
+
+            if [ "${label}" != "${label2}" ]; then
+                continue
+            fi
+
+            if [ "${start}" -ne "${start2}" ] || \
+               [ "${size}" -ne "${size2}" ]; then
```
oliv3r: ah so we ARE using the size, so what if it is not an integer ;) (we don't have a test for this do we? under the 'missformed/wrong table test)

scripts/prepare_disk.sh
```diff
@@ -128,6 +176,16 @@ if [ ! -r "${PARTITION_TABLE_FILE}" ]; then
     exit 1
 fi
 
+if [ ! -b "${TARGET_DISK}" ]; then
+    echo "Error, block device '${TARGET_DISK}' does not exist."
+    exit 1
+fi
+
+if ! is_resize_needed; then
```
oliv3r: be very wary that this will 'break' set -e

scripts/prepare_disk.sh
```diff
@@ -128,6 +176,16 @@ if [ ! -r "${PARTITION_TABLE_FILE}" ]; then
     exit 1
 fi
 
+if [ ! -b "${TARGET_DISK}" ]; then
+    echo "Error, block device '${TARGET_DISK}' does not exist."
+    exit 1
+fi
```
oliv3r, line 237: nice catch, i'm pretty sure at some point this was there, i must have miss-merged it at some point ...

test/buildenv.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
-
+#
+# Copyright (C) 2018 Ultimaker B.V.
+# Copyright (C) 2018 Raymond Siudak <raysiudak@gmail.com>
+# Copyright (C) 2018 Olliver Schinagl <oliver@schinagl.nl>
+#
+# SPDX-License-Identifier: AGPL-3.0+
 set -eu
```
oliv3r: :+1: on the license addition, but missing whitespace

(I'm thinking now that you have done this on EMP-302 or added the changes to this PR as well. Both are fine, but then you'll have to pull in the rework from the EMP-302 on these things too, if you want to merge those with this branch)

test/toolbox_image.sh
```diff
@@ -51,10 +57,11 @@ ________________________________________________________________________________
...
     LOOP_STORAGE_DEVICE="$(losetup --show --find --partscan "${rootfs_dir}${STORAGE_DEVICE_IMG}")"
 
-    mkfs.ext4 -q "${LOOP_STORAGE_DEVICE}p1"
-    mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p2"
-    mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p3"
+    mkfs.ext4 -q "${LOOP_STORAGE_DEVICE}p1" && fsck -y "${LOOP_STORAGE_DEVICE}p1"
+    mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p2" && fsck "${LOOP_STORAGE_DEVICE}p2"
+    mkfs.f2fs -q "${LOOP_STORAGE_DEVICE}p3" && fsck "${LOOP_STORAGE_DEVICE}p3"
 
+    sfdisk -d "${LOOP_STORAGE_DEVICE}"
```
oliv3r: yep these are deffinatly from EMP-302, i'd keep these changes to the EMP-302 branch. i'm thus not gonna look below this change for that reason, unless you can tell me it was intended ;)

oliv3r (changes requested):

test/toolbox_image.sh
```diff
@@ -193,7 +193,7 @@ cleanup()
 
 run_test()
 {
-    setup
+    setup &> /dev/null
```
oliv3r: while I generally approve (even part of this pull request)

for the future, maybe we need to add a variable at the top of our scripts
`UM_DEBUG="${UM_DEBUG:-false}"`

and then do things as:
```sh
SETUP_OUTPUT="&> /dev/null"
if "${UM_DEBUG}"; then
SETUP_OUTPUT=
fi
or something to make it configurable. Reason is, so that we can do pipeline reruns with these variables set/unset allowing us to debug steps on the CI end more easily ...

**Review by oliv3r — CHANGES_REQUESTED** [review]

File: build_for_ultimaker.sh (hunk 1 above)
Comment (raysiudak) [reply build_for_ultimaker.sh]: Yeah is see that changes from 305 are also in here because i rebased. I undid this and the changes should be in 302, so ignoring for now.

File: test/buildenv.sh (hunk 12 above)
Comment (raysiudak) [reply test/buildenv.sh]: Yeah part of EMP-302

File: scripts/prepare_disk.sh (hunk 10 above)
Comment (oliv3r) [reply scripts/prepare_disk.sh]: So we have to be very sparing in using this construct. (I'm not sure what happens if you call set -e in the function itself again, but that's more of a patchwork).

In this function, I think we are good as we pretty much check each and every call, except the cleanup unlink.

The only solution I can come up with however, is:
```sh
resize_needed="$(is_resize_needed; echo "${?}")"
if [ "${resize_needed}" -eq 1 ]; then
    echo "Partition resize not required."
    exit 0
fi
```
Which also looks a little quirky. But I suppose we cannot get around quirks with the tools we have right now.

oliv3r (dismissed):

oliv3r (dismissed):

scripts/prepare_disk.sh
```diff
@@ -22,6 +22,58 @@ usage()
...
+    test "${1}" -eq "${1}" 2> /dev/null
+}
+
+is_comment()
+{
+    test -z "${1%%#*}"
+}
+
+# A bit counter intuitive, a function can either return an exit code,
+# or set a global variable, or echo something, in this case the
```
daid: This sounds a bit odd, "it can set an exit code or a global". How is the expected use? And what global does it set, and for what reason? Or is that a side-effect that should be ignored?

scripts/prepare_disk.sh
```diff
@@ -22,6 +22,58 @@ usage()
...
+            fi
+
+            if [ "${table_label}" != "${disk_label}" ]; then
+                continue
+            fi
+
+            if [ "${table_start}" -ne "${disk_start}" ] || \
+               [ "${table_size}" -ne "${disk_size}" ]; then
+                changes_needed=true
+                return # returns from both loops!
```
daid: Shouldn't this be "return 0" to indicate that changes are needed? As as far as I can see this returns from both loops and the function. And thus setting changes_needed has no real effect within the function.

oliv3r (dismissed):

scripts/prepare_disk.sh
oliv3r: The manual states something like this. Shell functions cannot return values. After they finish, they just stop being. Using return, sets the exit value just before the function stops existing. You do not want to use exit, as that will exit the entire script.

One way to 'return' a value from a function, is to have it echo something, and capture the output when calling it. `OUTPUT=function` Whatever function outputs will end up in OUTPUT. It is a hack to have functions return values.

The other option to have a function return something is to set a (global, all variables are global in shell) variable.

This is standard shell behavior, and I believe is what raymond tried to explain with this comment for passers by not immediately understanding it.

scripts/prepare_disk.sh
oliv3r: You'd have to double check with raymond, but he found it did not. return only exists the subshell the while loop is being run under, as such, return should just return whatever was the last exit value within this subshell.

scripts/prepare_disk.sh
```diff
@@ -22,6 +22,58 @@ usage()
...
+}
+
+is_comment()
+{
+    test -z "${1%%#*}"
+}
+
+# A bit counter intuitive, a function can either return an exit code,
+# or set a global variable, or echo something, in this case the
+# exit code is 0 if a resize is required.
```
oliv3r: This I guess is over-explaining things a little, in that '0 is ok, anything else is error' and thus when reading `if is_resize_needed` the required exit value is counter intuitive as 1, rather then 0 to prevent having to read `if ! is_resize_needed`

Alternatively we could have used `no_resize_needed`, which would have resulted in the more expected exit value of 0 to indicate success of a function, lessening the (imo)slightly confusing comment.

Which in hind-sight, I think is more-readable.

scripts/prepare_disk.sh
raysiudak: Yess that is correct, but maybe i should explain it a bit better.

oliv3r: was the while that was executed in the subshell or was it the earlier version where you had
```sh
sfdisk | while ...
```
which causes the sfdisk to be called in an sfdisk.

As for explaining it better, be careful not to overexplain it, some of the things are knowing how to write shell, be careful not to explain shell scripting basics (even though it is VERY tempting).

scripts/prepare_disk.sh
raysiudak: Although it works, it does not look nice, i definitely agree. Nonetheless i decided to validate my finding once more it seems that i put myself on the wrong track before, probably with the 'sfdisk | while ...' construct, triggering an extra subshell. Much better now :P

scripts/prepare_disk.sh
```diff
@@ -32,6 +32,40 @@ is_comment()
     test -z "${1%%#*}"
 }
 
+# Returns 0 when resize is needed and 1 if not needed.
```
oliv3r: are you sure?
I think what you wanted, was `if is_resize_needed(true)`; then resize` but you are writing in the comment now, `if is_resize_needed(false); then resize`, which matches more with no_resize_needed, the return values haven't changed, so your comment does not your implementation anymore?

(If you do want 0 for resize, and 1 for not needed, might as well rename the function to no_resize_needed and drop the comment :)

scripts/prepare_disk.sh
```diff
@@ -32,6 +32,40 @@ is_comment()
...
+            fi
+
+            if [ "${table_label}" != "${disk_label}" ]; then
+                continue
+            fi
+
+            if [ "${table_start}" -ne "${disk_start}" ] || \
+               [ "${table_size}" -ne "${disk_size}" ]; then
+                unlink "${current_partition_table_file}"
+                return 0
```
oliv3r, line 60: so the while subshell was trigged by the sfdisk | while pipeline, makes sense.

To bad you can't use break && break or something to force both loops to break, removing the need for the current partition file.

Alternativly, trap EXIT with a cleanup function, that removes any existing partition table files and you can remove both unlink calls here, ensuring it is _always_ cleaned up. now, there's a chance of lingering files.

scripts/prepare_disk.sh
```diff
@@ -202,6 +236,12 @@ if [ ! -b "${TARGET_DISK}" ]; then
     exit 1
 fi
 
+resize_needed="$(is_resize_needed; echo "${?}")"
+if [ "${resize_needed}" -eq 1 ]; then
```
oliv3r: remember to update this depending on Your choice above. e.g.

```sh
resize_needed="$(no_resize_needed; echo "${?}")"
if [ ! "${resize_needed}" ]; then
```

scripts/prepare_disk.sh
```diff
@@ -196,6 +231,17 @@ if [ ! -r "${PARTITION_TABLE_FILE}" ]; then
     exit 1
 fi
 
+if [ ! -b "${TARGET_DISK}" ]; then
+    echo "Error, block device '${TARGET_DISK}' does not exist."
+    exit 1
+fi
+
+resize_needed="$(is_resize_needed; echo "${?}")"
```
oliv3r, line 239: also note, that I merely meant it as a suggestion that we should discuss. While this works, what happens if is_resize_needed produces output? then we'll have 'output + echo 0'. granted, for the if test it does not matter, as everything not 1, will fail and trigger a resize.

alternativly we can also force the output to be really just the exit value; but i think it's a little uglier?
```sh
if [ "${result}" -ne 0 ]; then
	echo "nok"
else
	echo "ok"
fi
```

scripts/prepare_disk.sh
raysiudak: Yes i am sure and this is the counter intuitive part of it. Is is not  if is_resize_needed(false/true) but if is_resize_needed(ok/nok)

scripts/prepare_disk.sh
raysiudak: Yeah it was triggered by the pipe and it would be nice to have breaks. 
The partition files are created in temp, so they are deleted.
i prefer it like so, without trap function.

scripts/prepare_disk.sh
raysiudak: Yes did and tested it

scripts/prepare_disk.sh
daid: Ah, oliver, you are mistaken here. A return value of "0" is "true" in shell scripting. (yes, confusing)

oliv3r: 0 being true is quite common in most *nix applications actually. U-Boot, busybox, the kernel, they all use 0 as 'good' return value as well.

But no worry, I'm fully aware and far from mistaken.

raysiudak: Ok guys, does it mean the reaview can now be approved :P

scripts/prepare_disk.sh
```diff
@@ -196,6 +231,17 @@ if [ ! -r "${PARTITION_TABLE_FILE}" ]; then
...
+if [ "${resize_needed}" -eq 1 ]; then
```
daid, line 240: with the latest changes, can't this just be?
```
if ! is_resize_needed; then
```

daid (approved):
One small possible improvement, otherwise fine.

scripts/prepare_disk.sh
oliv3r: only if resize_need = /bin/true; otherwise it will try to execute '1' which of course doesn't exist
which becomes worse if you do ! 1, which will always be true

daid: Read again, I proposed to change BOTH lines. Just like how is_integer and is_comment are used.

PR #22, opened 2018-08-24
EMP-378 Format partitions to honor the new disk size.

oliv3r:
Add a function that checks the partition layout and resizes partitions
where needed or reformats them if resize is not possible.

This pulls in blkid as we need to identify the partition type to be
able to call the correct tools.

This script does make the assumption that all partitions are f2fs
based, except for the BOOT partition, which is ext4 until u-boot
supports f2fs.

Addresses EMP-272, closes EMP-378

Signed-off-by: Olliver Schinagl <oliver@schinagl.nl>

### DIFF
--- a/scripts/prepare_disk.sh
+++ b/scripts/prepare_disk.sh
@@ -40,6 +40,61 @@ partition_sync()
     return 1
 }
 
+partitions_format()
+{
+    # Parse the output of sfdisk and temporally expand the Input Field Separator
+    # with ':=,' and treat them as whitespaces, in other words, ignore them.
+    sfdisk --quiet -d "${TARGET_DISK}" | \
+    while IFS="${IFS}:=," read -r disk_label _ disk_start _ disk_size _; do
+        while IFS="${IFS}:=," read -r table_label _ table_start _ table_size _; do
+            if [ -z "${disk_start}" ] || [ -z "${table_start}" ] || \
+               [ "${disk_start}" != "${table_start}" ]; then
+                continue
+            fi
+
+            if grep -q "${disk_label}" /proc/mounts; then
+                umount "${disk_label}"
+            fi
+
+            # Get the partition number from the label. e.g. /dev/loop0p1 -> p1
+	    # by grouping p with 1 or more digits and only printing the matchh
+            # and then format the partition. If the partition was already valid,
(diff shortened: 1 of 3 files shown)


scripts/prepare_disk.sh
```diff
@@ -40,6 +40,60 @@ partition_sync()
     return 1
 }
 
+partitions_format()
+{
+    # Parse the output of sfdisk and temporally expand the Input Field Separator
+    # with ':=,' and treat them as whitespaces, in other words, ignore them.
+    sfdisk --quiet -d "${TARGET_DISK}" |
```
raysiudak: Does this work without preceding '\' for continuation on the next line?

scripts/prepare_disk.sh
```diff
@@ -40,6 +40,60 @@ partition_sync()
...
+    while IFS="${IFS}:=," read -r disk_label _ disk_start _ disk_size _; do
```
raysiudak: i like disk_ and table_ naming

scripts/prepare_disk.sh
```diff
@@ -40,6 +40,60 @@ partition_sync()
...
+               [ "${disk_start}" != "${table_start}" ]; then
+                continue
+            fi
+
+            if grep -q "${disk_label}" /proc/mounts; then
+                umount "${disk_label}"
+            fi
+
+            # Get the partition number from the label. e.g. /dev/loop0p1 -> p1
+            # And then format the partition. If the partition was already valid,
```
raysiudak: And should be lowercase

scripts/prepare_disk.sh
```diff
@@ -40,6 +40,60 @@ partition_sync()
...
+            # just resize the existing one. If fsck or resize fails, reformat.
+            partition="$(echo "${disk_label}" | sed -rn 's/.*(p[[:digit:]]+$)/\1/p')"
```
raysiudak: i know 'vakmanschap' but please explain what is going on in the sed command

scripts/prepare_disk.sh
oliv3r: apparantly so, not quite sure as to why. The extra blank line may be gobbled up by the IFS is my guess. But i've fixed it.

raysiudak (approved):

PR #27, opened 2018-08-31
WIP EMP-388 Update um-update-toolbox to create a debian package.

daid:
This adds 3 extra tool requirements to the Docker container. As we need dpkg-deb and envsubst. And dpkg-deb does not work with busybox tar.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,9 +13,12 @@ LABEL Maintainer="software-embedded-platform@ultimaker.com" \
 RUN apk add --no-cache \
         e2fsprogs \
         f2fs-tools \
+        dpkg \
+        gettext \
         sfdisk \
         squashfs-tools \
         util-linux \
+        tar \
         xz \
     && \
     rm -f /var/cache/apk/*
--- a/build.sh
+++ b/build.sh
@@ -19,6 +19,12 @@ ROOTFS_DIR="${BUILD_DIR}/rootfs"
 
 COMMANDS="blkid busybox e2fsprogs-extra f2fs-tools rsync sfdisk"
 
+# Debian package information
+PREFIX="/usr/share/um-update-toolbox"
+RELEASE_VERSION="${RELEASE_VERSION:-9999.99.99}"
+PACKAGE_NAME="${PACKAGE_NAME:-um-update-toolbox-armhf}"
+DEB_PACKAGE="${PACKAGE_NAME}-${RELEASE_VERSION}.deb"
+
 cleanup()
 {
     mounts="$(grep "${ROOTFS_DIR}" /proc/mounts || true)"
@@ -113,6 +119,19 @@ compress_rootfs()
     echo "Created ${BUILD_DIR}/${TOOLBOX_IMAGE}."
 }
 
+create_debian_package()
+{
+    deb_dir="${BUILD_DIR}/debian_deb_build"
+
+    mkdir -p "${deb_dir}/DEBIAN"
+    RELEASE_VERSION="${RELEASE_VERSION}" PACKAGE_NAME="${PACKAGE_NAME}" envsubst "\${RELEASE_VERSION} \${PACKAGE_NAME}" < "${CUR_DIR}/debian/control.in" > "${deb_dir}/DEBIAN/control"
+
+    mkdir -p "${deb_dir}${INSTALL_DIR}"
+    cp "${BUILD_DIR}/${TOOLBOX_IMAGE}" "${deb_dir}${INSTALL_DIR}/"
+
+    dpkg-deb --build "${deb_dir}" "${BUILD_DIR}/${DEB_PACKAGE}"
+}
+
 usage()
 {
 cat <<-EOT
@@ -160,5 +179,6 @@ bootstrap_rootfs
 add_update_scripts
 bootstrap_unprepare
 compress_rootfs
+create_debian_package
 
 exit 0
--- a/build.sh
+++ b/build.sh
@@ -20,10 +20,11 @@ ROOTFS_DIR="${BUILD_DIR}/rootfs"
 COMMANDS="blkid busybox e2fsprogs-extra f2fs-tools rsync sfdisk"
 
 # Debian package information
-PREFIX="/usr/share/um-update-toolbox"
+PREFIX="${PREFIX:-/usr}"
+PACKAGE_NAME="${PACKAGE_NAME:-um-update-toolbox}"
+INSTALL_DIR="${INSTALL_DIR:-${PREFIX}/share/${PACKAGE_NAME}}"
 RELEASE_VERSION="${RELEASE_VERSION:-9999.99.99}"
-PACKAGE_NAME="${PACKAGE_NAME:-um-update-toolbox-armhf}"
-DEB_PACKAGE="${PACKAGE_NAME}-${RELEASE_VERSION}.deb"
+DEB_PACKAGE="${PACKAGE_NAME}_${ARCH}-${RELEASE_VERSION}.deb"
 
 cleanup()
 {
--- a/build_for_ultimaker.sh
+++ b/build_for_ultimaker.sh
@@ -87,7 +87,7 @@ run_build()
         ./build.sh
     fi
     # We need to place the .deb file in the root folder, as jedi-build expects it there.
-    cp ${BUILD_DIR}/*.deb "./"
+    cp "${BUILD_DIR}/"*.deb "./"
 }
 
 run_tests()
--- a/build_for_ultimaker.sh
+++ b/build_for_ultimaker.sh
@@ -86,6 +86,7 @@ run_build()
     else
         ./build.sh
     fi
+
     # We need to place the .deb file in the root folder, as jedi-build expects it there.
     cp "${BUILD_DIR}/"*.deb "./"
 }
--- a/build_for_ultimaker.sh
+++ b/build_for_ultimaker.sh
@@ -86,6 +86,8 @@ run_build()
     else
         ./build.sh
     fi
+    # We need to place the .deb file in the root folder, as jedi-build expects it there.
+    cp ${BUILD_DIR}/*.deb "./"
 }
 
 run_tests()
--- a/debian_control
+++ b/debian_control
@@ -1,0 +1,6 @@
+Package: ${PACKAGE_NAME}
+Version: ${RELEASE_VERSION}
+Architecture: armhf
+Maintainer: Ultimaker <info@ultimaker.com>
+Priority: optional
+Description: Toolbox used during the system update.
--- a/test/buildenv.sh
+++ b/test/buildenv.sh
@@ -12,7 +12,7 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
 BINFMT_MISC="${BINFMT_MISC:-/proc/sys/fs/binfmt_misc/}"
 
 FILESYSTEMS="ext4 f2fs overlay squashfs tmpfs"
-COMMANDS="apk blkid fsck.ext4 fsck.f2fs losetup mkfs.ext4 mkfs.f2fs mksquashfs sed sfdisk shuf xz"
+COMMANDS="apk blkid dpkg fsck.ext4 fsck.f2fs gettext losetup mkfs.ext4 mkfs.f2fs mksquashfs sed sfdisk shuf tar xz"
 
 ARMv7_MAGIC="7f454c4601010100000000000000000002002800"
 
--- a/debian/control.in
+++ b/debian/control.in
@@ -1,6 +1,6 @@
 Package: ${PACKAGE_NAME}
 Version: ${RELEASE_VERSION}
 Architecture: armhf
-Maintainer: Ultimaker <info@ultimaker.com>
+Maintainer: Ultimaker <software-embedded-platform@ultimaker.com>
 Priority: optional
 Description: Toolbox used during the system update.


Dockerfile
```diff
@@ -17,6 +17,9 @@ RUN apk add --no-cache \
         squashfs-tools \
         util-linux \
         xz \
+        dpkg \
+        gettext \
+        tar \
```
oliv3r: please keep the list sorted

build_for_ultimaker.sh
```diff
@@ -86,6 +86,8 @@ run_build()
     else
         ./build.sh
     fi
+    # We need to place the .deb file in the root folder, as jedi-build expects it there.
```
oliv3r: replace the comment with a single new line

we don't have to add this comment to each package that we build, it's an interface between a package and jedi-build and thus it should be documented somewhere more appropriate, rather then in each build script.

debian_control
```diff
@@ -0,0 +1,6 @@
+Package: ${PACKAGE_NAME}
+Version: ${RELEASE_VERSION}
+Architecture: armhf
+Maintainer: Ultimaker <info@ultimaker.com>
```
oliv3r: `Maintaner: Ultimaker <software-embedded-platform@ultimaker.com>`

debian_control
```diff
@@ -0,0 +1,6 @@
+Package: ${PACKAGE_NAME}
```
oliv3r: please but the file under the debian subdirectory as just control, as per standard debian packaging policy. Since this is an 'input' file to a replace stuff with, give it the '.in' suffix.

build.sh
```diff
@@ -113,6 +119,24 @@ compress_rootfs()
     echo "Created ${BUILD_DIR}/${TOOLBOX_IMAGE}."
 }
 
+create_debian_package()
+{
+    DEB_DIR=$(mktemp -d -t)
```
oliv3r: -t appears to be deprecated, and your not using a template? Also create DEB_DIR under "${BUILD_DIR}"

build.sh
```diff
@@ -113,6 +119,24 @@ compress_rootfs()
...
+
+    # Create a debian control file to pack up a debian package
+    mkdir -p "${DEB_DIR}/DEBIAN"
```
oliv3r: Instead, I recommend to put generated debian files under the `mkdir -p "${BUILD_DIR}/${DEB_DIR}/debian"` instead as it will be cleaned up by the build script whenever BUILD_DIR gets cleaned up.

p.s. while both are used, debian seems to be more common then DEBIAN

build.sh
```diff
@@ -113,6 +119,24 @@ compress_rootfs()
     echo "Created ${BUILD_DIR}/${TOOLBOX_IMAGE}."
 }
 
+create_debian_package()
+{
+    DEB_DIR=$(mktemp -d -t)
+
+    # Create a debian control file to pack up a debian package
```
oliv3r: i think the function name documents this more then the comment does

debian_control
```diff
@@ -0,0 +1,6 @@
...
+Version: ${RELEASE_VERSION}
```
oliv3r: you are abusing environment variables here to work with envsubst, rather, do it the CMake/autoconf way, as that is what you are really doing here:
`Version: @RELEASE_VERSION@`
which is far more recognizable ontop of that

daid:
Note: CI fails due to Dockerfile changes.

debian_control
```diff
@@ -0,0 +1,6 @@
...
+Architecture: armhf
+Maintainer: Ultimaker <info@ultimaker.com>
+Priority: optional
```
oliv3r: missing 'Section'
I recommend
`Section: utils`

build.sh
```diff
@@ -113,6 +119,24 @@ compress_rootfs()
...
+    mkdir -p "${DEB_DIR}/DEBIAN"
+    RELEASE_VERSION="${RELEASE_VERSION}" PACKAGE_NAME="${PACKAGE_NAME}" envsubst "\${RELEASE_VERSION} \${PACKAGE_NAME}" < "${CUR_DIR}/debian_control" > "${DEB_DIR}/DEBIAN/control"
```
oliv3r: You are adding envsubst where we don't really need it.

I think we can happily use:
```sh
sed -e 's/@RELEASE_VERSION@/'"${RELEASE_VERSION}"'/g \
    -e 's/@PACKAGE_NAME@/'"${PACKAGE_NAME}"'/g' \
   "${CUR_DIR}/debian/control.in" > "${BUILD_DIR}${DEB_DIR}/debian/control"
```

build.sh
```diff
@@ -19,6 +19,12 @@ ROOTFS_DIR="${BUILD_DIR}/rootfs"
 
 COMMANDS="blkid busybox e2fsprogs-extra f2fs-tools rsync sfdisk"
 
+# Debian package information
+PREFIX="/usr/share/um-update-toolbox"
```
oliv3r: Could you add 'DATAROOTDIR="${DATAROOTDIR:-/usr/share/}" for the first half, and 'PACKAGE_NAME="um-update-toolbox" for the second half.
As for the complete helper (PREFIX) I think it's clearer to not use it.

build.sh
```diff
@@ -113,6 +119,24 @@ compress_rootfs()
...
 
+create_debian_package()
+{
+    DEB_DIR=$(mktemp -d -t)
+
+    # Create a debian control file to pack up a debian package
+    mkdir -p "${DEB_DIR}/DEBIAN"
+    RELEASE_VERSION="${RELEASE_VERSION}" PACKAGE_NAME="${PACKAGE_NAME}" envsubst "\${RELEASE_VERSION} \${PACKAGE_NAME}" < "${CUR_DIR}/debian_control" > "${DEB_DIR}/DEBIAN/control"
+
+    mkdir -p "${DEB_DIR}${PREFIX}"
```
oliv3r: I think this becomes more clearer what the intend is if it reads: `mkdir -p "${BUILD_DIR}${DEB_DIR}${DATAROOTDIR}${PROJECT_NAME}"

build.sh
```diff
@@ -113,6 +119,24 @@ compress_rootfs()
...
+    cp "${BUILD_DIR}/${TOOLBOX_IMAGE}" "${DEB_DIR}${PREFIX}/"
+
+    # Build the actual deb file.
```
oliv3r: remove comment, it's stating the obvious

build.sh
```diff
@@ -113,6 +119,24 @@ compress_rootfs()
...
+    dpkg-deb --build "${DEB_DIR}" "${BUILD_DIR}/${DEB_PACKAGE}"
+
+    # Clean up after ourselves
+    rm -rf "${DEB_DIR}"
```
oliv3r: all cleanup should be done in the cleanup routine where appropriate, as , also, never just rm -rf ${VAR} without doing some check first. (e.g. is length greater then 1 to avoid /, test -d if it is a dir etc. But by letting the cleanup step handling the cleanup your all set anyway.

build.sh
```diff
@@ -113,6 +119,24 @@ compress_rootfs()
...
+
+    # Create a debian control file to pack up a debian package
+    mkdir -p "${DEB_DIR}/DEBIAN"
+    RELEASE_VERSION="${RELEASE_VERSION}" PACKAGE_NAME="${PACKAGE_NAME}" envsubst "\${RELEASE_VERSION} \${PACKAGE_NAME}" < "${CUR_DIR}/debian_control" > "${DEB_DIR}/DEBIAN/control"
+
+    mkdir -p "${DEB_DIR}${PREFIX}"
+    cp "${BUILD_DIR}/${TOOLBOX_IMAGE}" "${DEB_DIR}${PREFIX}/"
+
+    # Build the actual deb file.
+    dpkg-deb --build "${DEB_DIR}" "${BUILD_DIR}/${DEB_PACKAGE}"
```
oliv3r: build up the package name here, as ${PACAKGE_NAME}-${PACKAGE_VERSION}_${ARCH}.deb. rather then hiding it in helper vars.

build.sh
```diff
@@ -19,6 +19,12 @@ ROOTFS_DIR="${BUILD_DIR}/rootfs"
 
 COMMANDS="blkid busybox e2fsprogs-extra f2fs-tools rsync sfdisk"
 
+# Debian package information
+PREFIX="/usr/share/um-update-toolbox"
+RELEASE_VERSION="${RELEASE_VERSION:-9999.99.99}"
+PACKAGE_NAME="${PACKAGE_NAME:-um-update-toolbox-armhf}"
```
oliv3r: drop the hardcoded architecture name from here

oliv3r (dismissed):
Also please add tests to the buildenv.sh script to test for the new packages. Especially in the case for tar, ensure that the installed tar does have the needed options, that busybox tar is lacking, to ensure the test fails with just busybox tar.

also could you please rename your branch to EMP*, to ensure JIRA doesn't get thrown off.

Dockerfile
daid: I did, it's sorted by order of requirement. (AKA, you are making up style rules as you go)

build.sh
daid: You introduced it in the repo name.

build.sh
daid: "and your not using a template" Me confused ?

build.sh
daid: was copied from um-kernel: https://github.com/Ultimaker/um-kernel/blob/master/build_for_ultimaker.sh#L297

build.sh
daid: Copied from um-kernel, and I fully disagree, sed is ugly and error prone.

build.sh
daid: Copied from um-kernel.

build.sh
daid: First part ok.

Second part. Not ok, false sense of security. And even wrong. ```rm -rf ""`` does nothing.

build_for_ultimaker.sh
daid: Except that every single other package builds the deb file "in place", where this package cannot due to docker. So it's different then everything else, and I documented the reason.

debian_control
daid: Copied from um-kernel.

debian_control
daid: Section is optional.

debian_control
daid: I disagree, and once again, copied from um-kernel.

debian_control
daid: And the less things we put in here that we don't really need. The less we need to maintain and think about.

build.sh
raysiudak: It is an extra docker dependency to gettext that is not really needed. 
If you decide not to change then resolve conversation pls

debian_control
daid: Done

build.sh
daid: Copied upper case DEBIAN from um-kernel. The deletion was removed as it's part of the build folder now.

raysiudak (approved):

test/buildenv.sh
```diff
@@ -12,7 +12,7 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
 BINFMT_MISC="${BINFMT_MISC:-/proc/sys/fs/binfmt_misc/}"
 
 FILESYSTEMS="ext4 f2fs overlay squashfs tmpfs"
-COMMANDS="apk blkid fsck.ext4 fsck.f2fs losetup mkfs.ext4 mkfs.f2fs mksquashfs sed sfdisk shuf xz"
+COMMANDS="apk blkid dpkg fsck.ext4 fsck.f2fs gettext losetup mkfs.ext4 mkfs.f2fs mksquashfs sed sfdisk shuf tar xz"
```
oliv3r, line 15: Don't test for the gettext command, test for the 'envsubst' command, as that is the command that you need.

Also, no test was added for the missing features of tar that dpkg 'depends' on. Tar will always be available as it's part of busybox.

Dockerfile
oliv3r: David, there is no need to be rude. These are rules we agreed upon within the platform-team, and try to follow up until now, check earlier review comments etc. Also, it's common docker style. Reason is, that you can more easily spot duplicates etc.

Also, you did not, you did not sort it in order of requirement, as dpkg requires tar, and thus those should have been together.

In any case, we do not have a rule 'sort in order of requirement'.

build.sh
oliv3r: [The -t parameter expects a template](https://linux.die.net/man/1/mktemp),
`
-t
    interpret TEMPLATE as a single file name component, relative to a directory: $TMPDIR, if set; else the directory specified via -p; else /tmp [deprecated]
`
In other words, the correct usage is: `mktemp -t mytemplateXXXXXX`. By not supplying the template to the -t parameter, you are using the deprecated /tmp case.

build.sh
oliv3r: yeah the um-kernel build script needs some major cleanups as it is the old build script we have been using for the past years. We thus should use this opportunity to improve ourselves, rather then hide behind the 'we did it like this before so it is ok to just copy it. :)

build.sh
oliv3r: also here, the um-kernel build script needs some major cleanups as it is the old build script we have been using for the past years. We thus should use this opportunity to improve ourselves, rather then hide behind the 'we did it like this before so it is ok to just copy it. :)

debian_control
oliv3r: once more here, the um-kernel build script needs some major cleanups as it is the old build script we have been using for the past years. We thus should use this opportunity to improve ourselves, rather then hide behind the 'we did it like this before so it is ok to just copy it. :)

build.sh
oliv3r: your disagreement is noted, but we are using sed all over already, and are not using envsubst.

I fully disagree with envsubst, it's ugly and unneeded.

build.sh
oliv3r: also here, the um-kernel build script needs some major cleanups as it is the old build script we have been using for the past years. We thus should use this opportunity to improve ourselves, rather then hide behind the 'we did it like this before so it is ok to just copy it. :)

build.sh
oliv3r: but rm -rf "/" does everything, so it's not a false sense of security, it's just adding some basic checks to avoid the most obvious mistakes.

build.sh
oliv3r: that is absolutely true, but avoiding hardcoded variables is always better, In any case, i suggested the dropping of the hardcoded arch, as it should not be part of the package name. I wrongfully assumed that my intention was obvious from the previous comment

test/buildenv.sh
daid: Busybox tar won't work for dpkg, so the dpkg package incorrect doesn't depend on gnu-tar, but needs it.

Dockerfile
daid: > These are rules we agreed upon within the platform-team

Write them down, as I'm in the team, and not aware of the rules. So there are no rules. Hint, first thing when I get back from vacation is asking if you want to write code style rules for shell scripting. As you clearly have a lot, but they are not written down.

build.sh
daid: rm -rf "/" does nothing, as rm protects against that. And you asked for an empty check, and thus the check does not protect against what you suggest.

rm -rf "/home" does major damage, and then isn't protected. Hench, false sense of security.

debian_control
daid: (note, added the copy from lines to indicate this also needs fixing in um-kernel. I did fix those things)

build.sh
daid: sed with user supplied variables is just asking for problems.

PR #31, opened 2018-09-03
Emp 395 isolate config files

raysiudak:
This branch is based on EMP-311, merge that one first.


.gitlab-ci.yml
```diff
@@ -54,6 +59,11 @@ lint_package_build_script:
   script:
     - shellcheck -C -f tty -s sh build.sh
 
+lint_prepare_disk_script:
+  <<: *shell_linting_common
+  script:
+    - shellcheck -C -f tty -s sh scripts/30_prepare_disk.sh
```
daid: General comment, wouldn't it make more sense to just "glob" the shell files, and thus have the CI always check all shell files, instead of needing to maintain a "shadow list"?

build.sh
```diff
@@ -64,21 +65,37 @@ bootstrap_unprepare()
...
+    local_script_dir="${CUR_DIR}/scripts"
     for script in "${local_script_dir}"/*.sh; do
         basename="${script##*/}"
         echo "Installing ${script} on '${target_script_dir}/${basename}'."
         cp "${script}" "${target_script_dir}/${basename}"
         chmod +x "${target_script_dir}/${basename}"
     done
+
+    entrypoint_script="start_update.sh"
+    chroot "${ROOTFS_DIR}" ln -s "${SYSTEM_UPDATE_DIR}.d/${entrypoint_script}" "/sbin/${entrypoint_script}"
```
daid, line 82: I'm not sure if it makes sense that the start_update.sh script is located in the .d directory. As .d directories are generally for glob actions, and you don't want the entry point to be picked up by a glob rule.

daid: So maybe make a split of glob files and other files?

raysiudak: Yes true, i will fix this in my next task if that is alright with you. In the next tast i will actually be globbing the scripts and executing them.

.gitlab-ci.yml
raysiudak: Yes good point, the shadow list is a maintenance burden. I will see if this works.

PR #34, opened 2018-09-06
Emp 390 integrate into start update

raysiudak:


.gitlab-ci.yml
```diff
@@ -49,6 +49,11 @@ lint_package_build_script:
   script:
     - shellcheck -C -f tty -s sh build.sh
 
+lint_prepare_disk_script:
```
daid: It's back again?

scripts/start_update.sh
```diff
@@ -5,8 +5,199 @@
...
+        return 1
+    fi
+
+    temp_folder="$(mktemp -d)"
+    if ! mount --bind "${temp_folder}" "${TOOLBOX_MOUNT}${UPDATE_ROOTFS_SOURCE}"; then
+        echo "Error, update failed: temp source update directory: '${TOOLBOX_MOUNT}${UPDATE_ROOTFS_SOURCE}' cannot be mounted."
+        return 1
+    fi
+
+    if ! tar xvfJ "${UPDATE_MOUNT}/${update_rootfs_archive}" -C "${TOOLBOX_MOUNT}${UPDATE_ROOTFS_SOURCE}" \
```
daid: The verbose flag on tar causes each filename to be printed to the serial console, which will slow down the extraction on the target. As the extraction will be faster then the serial port and buffers will run full.

daid: Wait, it's verbose, but the output is directed to /dev/null?

scripts/start_update.sh
```diff
@@ -5,8 +5,199 @@
...
+        rm "${UPDATE_MOUNT:?}/${update_rootfs_archive:?}"
+    fi
+
+    if grep -q "${TOOLBOX_MOUNT}${UPDATE_ROOTFS_SOURCE}" /proc/mounts; then
+        echo "Cleaning up, unmount: '${TOOLBOX_MOUNT}${UPDATE_ROOTFS_SOURCE}'."
+        umount "${TOOLBOX_MOUNT}${UPDATE_ROOTFS_SOURCE}"
+    fi
+}
+
+extract_update_rootfs()
```
daid, line 59: I kinda would have expected this to happen as 40_extract_rootfs.sh or as part of the 50_update_files.sh instead of outside the apply update scripts?

daid (approved):

PR #35, opened 2018-09-06
Backup data on partitions before resizing them, and restore the data …

daid:
…if the partition was wiped. This backs up to /tmp/, if tmpfs is full, then the partition data that does not fit will be ignored. EMP-401

### DIFF
--- a/scripts/30_prepare_disk.sh
+++ b/scripts/30_prepare_disk.sh
@@ -213,22 +213,27 @@ partition_resize()
 
 backup_data()
 {
-    temp_mount_dir=$(mktemp -d)
+    temp_mount_dir="$(mktemp -d)"
 
     sfdisk --quiet --dump "${TARGET_STORAGE_DEVICE}" | \
     while IFS="${IFS}:=," read -r disk_device _ disk_start _ disk_size _; do
         if [ -b "${disk_device}" ]; then
-            if grep -q "${disk_device}" /proc/mounts; then
+
+            if grep -q "${disk_device}" "/proc/mounts"; then
                 umount "${disk_device}"
             fi
+
             backup_file="/tmp/backup${disk_device}.tar.gz"
+
             if mount "${disk_device}" "${temp_mount_dir}"; then
                 mkdir -p "$(dirname "${backup_file}")"
+
                 echo "Backing up ${disk_device} as ${backup_file}"
-                if ! tar -czf "${backup_file}" -C "${temp_mount_dir}" .; then
+                if ! tar -czf "${backup_file}" -C "${temp_mount_dir}" . > /dev/null; then
                     echo "Backup failed, removing backup. Partition will be empty."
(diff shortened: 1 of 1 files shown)


scripts/30_prepare_disk.sh
```diff
@@ -208,6 +208,51 @@ partition_resize()
...
+    sfdisk --quiet --dump "${TARGET_STORAGE_DEVICE}" | \
+    while IFS="${IFS}:=," read -r disk_device _ disk_start _ disk_size _; do
+        if grep -q "${disk_device}" /proc/mounts; then
+            umount "${disk_device}"
+        fi
+        backup_file="/tmp/backup${disk_device}.tar.gz"
+        if mount "${disk_device}" "${temp_mount_dir}"; then
+            mkdir -p "$(dirname "${backup_location}")"
+            if ! tar -czf "${backup_location}" -C "${temp_mount_dir}" .; then
+                rm "${backup_location}"
```
raysiudak: I think you should redirect the output to /dev/null not to pollute the output data and echo that the backup for disk device partition x was successful. 
Same goes for restore.

scripts/30_prepare_disk.sh
```diff
@@ -208,6 +208,51 @@ partition_resize()
     sfdisk --quiet "${TARGET_STORAGE_DEVICE}" < "${SYSTEM_UPDATE_DIR}/${PARTITION_TABLE_FILE}"
 }
 
+backup_data()
+{
+    temp_mount_dir=$(mktemp -d)
+
+    sfdisk --quiet --dump "${TARGET_STORAGE_DEVICE}" | \
+    while IFS="${IFS}:=," read -r disk_device _ disk_start _ disk_size _; do
+        if grep -q "${disk_device}" /proc/mounts; then
```
raysiudak: Shellcheck won't complain, but i but it is more consistent to also add quotes to "/proc/mount", same goes for the restore.

raysiudak (changes requested):
In general, for the next version we should optimize it, not everything needs to be copied or restored. And then we could create a 25_backup.sh and a 35_restore.sh or 55_restore.sh script.

scripts/30_prepare_disk.sh
```diff
@@ -208,6 +208,51 @@ partition_resize()
...
+            umount "${disk_device}"
+        fi
+        backup_file="/tmp/backup${disk_device}.tar.gz"
+        if [ -f "${backup_file}" ]; then
+            mount "${disk_device}" "${temp_mount_dir}"
+            # We only restore if the parition looks empty. There can be a lost+found on an empty partition, so ignore that.
+            if [ "$(ls "${temp_mount_dir}" | wc -l)" < 2; then
```
raysiudak: Shellcheck SC2012: Use find instead of ls to better handle non-alphanumeric filenames.

raysiudak (approved):

PR #36, opened 2018-09-06
Emp 390 integrate into start update

raysiudak:
Apparently forgot to push the latest changes


scripts/start_update.sh
```diff
@@ -188,16 +212,18 @@ if [ ! -f "${TOOLBOX_MOUNT}${SYSTEM_UPDATE_DIR}/${UPDATE_EXCLUDE_LIST_FILE}" ];
...
-# sort scripts with globbing
-# execute all scripts in chroot
-cleanup
+if ! perform_update; then
+   echo "Update failed: unable to perform update."
+   exit 1
+fi
 
-echo "Ok, update success."
+cleanup
```
daid: Double cleanup on normal exit:
http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_12_02.html
```If a signal is 0 or EXIT, the COMMANDS are executed when the shell exits.```

File: scripts/start_update.sh [hunk 2]
```diff
@@ -102,6 +104,28 @@ extract_update_rootfs()
     return 0
```
daid, line 132: Shouldn't we unmount ${UPDATE_MOUNT} here? As we no longer need it, and if it's still mounted when we are going to resize we might run into issues. Not sure if the resize script can unmount it from the chroot if it is mounted outside the chroot...

scripts/start_update.sh
raysiudak: My bad

daid (approved):

PR #38, opened 2018-09-07
Fix 'could not remove rootfs' warning and some small improvements

raysiudak:
- The init.sh script from the um-kernel is in control of the update files
  and therefore should be the one to delete them.
- Add check for checksum file in the start_udpate script since it is an
  input.

Contributes to EMP-382

### DIFF
--- a/scripts/start_update.sh
+++ b/scripts/start_update.sh
@@ -45,11 +45,6 @@ cleanup()
         rm -rf "${TOOLBOX_MOUNT:?}/${UPDATE_ROOTFS_SOURCE:?}"/*
     fi
 
-    if [ -f "${UPDATE_MOUNT}/${update_rootfs_archive}" ]; then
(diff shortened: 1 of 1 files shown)


scripts/start_update.sh
```diff
@@ -236,6 +231,11 @@ if [ ! -f "${TOOLBOX_MOUNT}${SYSTEM_UPDATE_DIR}/${PARTITION_TABLE_FILE}" ]; then
     exit 1
 fi
 
+if [ ! -f "${TOOLBOX_MOUNT}${SYSTEM_UPDATE_DIR}/${PARTITION_TABLE_FILE}.sha512" ]; then
+    echo "Error, update failed: '${TOOLBOX_MOUNT}${TOOLBOX_MOUNT}${SYSTEM_UPDATE_DIR}/${PARTITION_TABLE_FILE}' not found."
```
daid: Message lacks ".sha512" postfix.

daid (changes requested):

daid (approved):

PR #41, opened 2018-09-12
Emp 417 cleanup toolbox image test

raysiudak:


.gitlab-ci.yml
```diff
@@ -32,7 +32,8 @@ stages:
 lint_test_scripts:
   <<: *shell_linting_common
   script:
-    - shellcheck -C -f tty -s sh test/*.sh
+    - shellcheck -x -C -f tty -s sh test/*.sh
+    - shellcheck -x -C -f tty -s sh test/include/*.sh
```
oliv3r: since your touching these lines, maybe do `"test/"*`? (e.g. quote the directory names)

test/include/armhf_emulation.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/sh
```
oliv3r: missing copyright/license header, i assume the rest of the file is identical to what we already had.

the files and function names however are quite irrelevant. This is the test_setup and test_teardown stuff, so name them as such. I'd put them all in their own file(s) tbh. test_helpers.sh or test/helpers/common.sh or something for example.
Now it is lying about what is going on.

test/include/armhf_emulation.sh
```diff
@@ -0,0 +1,102 @@
...
+    echo "Setup armhf emulation."
+
+    # devtmps needs to be re-mounted when running in docker
+    if ! grep -qE "/dev.*devtmpfs" "/proc/mounts"; then
+        mount -t "devtmpfs" "none" "/dev"
+        is_dev_setup_mounted=true
+    fi
+
+    # overlayfs is required because chroot expects the qemu binary in /bin
+    # and modified partition tables are created in this test
```
oliv3r: not sure if we had this coment here before, but it's not entirey true, we don't test anything, nor modify any partition tables here :p, also the kernel expects the arm emulation binary there, even in a chrooted environment.

test/toolbox_image.sh
```diff
@@ -12,243 +12,65 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
...
-update_mount=""
-
-result=0
-
-is_dev_setup_mounted=false
 exit_on_failure=false
+result=0
 
+# shellcheck source=test/include/armhf_emulation.sh
+. "test/include/armhf_emulation.sh"
```
oliv3r: I'd recommend putting includes at the very top of the file. there should not be any dependencies which will 'force' that by having it on top.

test/toolbox_image.sh
```diff
@@ -12,243 +12,65 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
...
-        mount -t devtmpfs none "/dev"
-        is_dev_setup_mounted=true
-    fi
-
-    mount -t devtmpfs none "${rootfs_dir}/dev"
-
-    dd if=/dev/zero of="${rootfs_dir}/${TMP_TEST_IMAGE_FILE}" bs=1 count=0 seek=128M 2> /dev/null
-
-    create_dummy_storage_device
+    set -e
```
oliv3r: missing -u
and why do you explicitly have to 're'-set, -e during setup?

test/toolbox_image.sh
```diff
@@ -12,243 +12,65 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
...
-        return
-    fi
-
-    for partition_table_file in "${rootfs_dir}${PARTITION_TABLE_FILE}"*; do
-        unlink "${partition_table_file}"
-    done
-
-    if [ -d "${rootfs_dir}${UPDATE_ROOTFS_SOURCE}" ]; then
-        rm -rf "${rootfs_dir}${UPDATE_ROOTFS_SOURCE}"
+    set -e
```
oliv3r: as above, and missing new line

test/toolbox_image.sh
```diff
@@ -12,243 +12,65 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
...
     fi
 
-    if [ -d "${update_mount}" ]; then
-        rm -r "${update_mount:?}"
-    fi
+    teardown_armhf_emulation "${toolbox_root_dir}"
 
-    if [ -b "${LOOP_STORAGE_DEVICE}" ]; then
-        losetup -d "${LOOP_STORAGE_DEVICE}"
+    if grep -qE "${toolbox_root_dir}" "/proc/mounts"; then
```
oliv3r: do we need the -E here to grep? i'm not seeing a regular expression so I guess no?

test/toolbox_image.sh
```diff
@@ -12,243 +12,65 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
...
+        umount "${toolbox_root_dir}"
     fi
 
-    if [ -f "${rootfs_dir}${STORAGE_DEVICE_IMG}" ]; then
-        unlink "${rootfs_dir}${STORAGE_DEVICE_IMG}"
+    if [ -d "${toolbox_root_dir}" ]; then
+        rm -rf "${toolbox_root_dir:?}"
```
oliv3r: i'm slightly uneasy with this command, as the ? does protect you from empty variable (rm -rf NULL exits also i believe) it does not protect from rm -rf "/".

we could say, we add the -t="um-toolbox.XXXXXX" parameter to mktemp and add the test `&& [ "${toolbox_root_dir#"um-toolbox.."}" != "um-toolbox." ]`

this ensures that we only ever delete the directory if it at least contains our directory template.

While i'd leave the -d check in place, that would be then optional as well, as we create something, we delete it.

test/toolbox_image.sh
```diff
@@ -12,243 +12,65 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
...
-    echo "      umount '/dev' \\"
-    echo "    fi && \\"
-    echo "    umount '${rootfs_dir}/proc' && \\"
-    echo "    umount '${rootfs_dir}' && \\"
-    echo "    rmdir '${rootfs_dir}' && \\"
-    echo "    umount '${overlayfs_dir}/rom' && \\"
-    echo "    rmdir '${overlayfs_dir}/rom' && \\"
-    echo "    umount '${overlayfs_dir}/' && \\"
-    echo "    rmdir '${overlayfs_dir}/'"
+    echo "    unlink '${toolbox_root_dir}/${TMP_TEST_IMAGE_FILE}'&& \\"
```
oliv3r: missing whitespace between ' and &&

test/toolbox_image.sh
```diff
@@ -12,243 +12,65 @@ ARM_EMU_BIN="${ARM_EMU_BIN:-}"
...
+    echo "    umount '${toolbox_root_dir}' && \\"
+    echo "    rm -rf '${toolbox_root_dir:?}' && \\"
```
oliv3r: I'm not sure if the ? is wise here

if toolbox_root_dir is NULL (the -u should protect you for the unset variable at the very start of this function/script) it will print:
unlink 'NULL'/tmp_test_image.img && \
umount 'NULL' && \
<exit 1>

test/toolbox_image.sh
```diff
@@ -288,203 +110,88 @@ run_test()
...
-    chroot "${rootfs_dir}" /usr/sbin/fsck.f2fs "${TMP_TEST_IMAGE_FILE}" 1> /dev/null
+    test_execute_mkfs_f2fs || return 1
+    chroot "${toolbox_root_dir}" /usr/sbin/resize.f2fs "${TMP_TEST_IMAGE_FILE}" || return 1
+    chroot "${toolbox_root_dir}" /usr/sbin/fsck.f2fs "${TMP_TEST_IMAGE_FILE}" || return 1
 }
 
 test_execute_mount()
 {
-   chroot "${rootfs_dir}" /bin/mount --version 1> /dev/null
+   chroot "${toolbox_root_dir}" /bin/mount --version
```
oliv3r: I think you want to add `|| return 1` here

test/toolbox_image.sh
```diff
@@ -288,203 +110,88 @@ run_test()
...
-    jedi_partition_table_file="jedi_emmc_sfdisk.table"
-    test -f "${rootfs_dir}${SYSTEM_UPDATE_DIR}/${jedi_partition_table_file}" || return 1
-}
-
-test_execute_disk_prepare_sha512_nok()
-{
-    chroot "${rootfs_dir}" sha512sum "${PARTITION_TABLE_FILE}" > "${rootfs_dir}${PARTITION_TABLE_FILE}.sha512"
-    echo "corrupted partition table data" >> "${rootfs_dir}${PARTITION_TABLE_FILE}"
-    chroot "${rootfs_dir}" "${PREPARE_DISK_COMMAND}" -t "${PARTITION_TABLE_FILE_NAME}" -d "${LOOP_STORAGE_DEVICE}" || return 0
+    chroot "${toolbox_root_dir}" /usr/bin/rsync --version
```
oliv3r: ditto

test/toolbox_image.sh
```diff
@@ -288,203 +110,88 @@ run_test()
...
-
-    rsync --exclude-from "${UPDATE_EXCLUDE_LIST_FILE}" -c -a -x --dry-run \
-        "${rootfs_dir}${UPDATE_ROOTFS_SOURCE}/" "${update_target}/" || return 1
-
-    umount "${update_target}"
-
-    if [ -z "${update_target##*/tmp/*}" ]; then
-        rm -rf "${update_target}"
-    fi
+    test -f "${toolbox_root_dir}${JEDI_EXCLUDE_LIST_FILE}" || return 1
```
oliv3r, line 179: test if the file is readable, just the mere existence is not enough

test/toolbox_image.sh
```diff
@@ -288,203 +110,88 @@ run_test()
...
 }
 
-test_execute_start_update_update_rootfs_mount_point_exists()
+test_jedi_partition_table_file_exists()
 {
-    update_rootfs_source="/mnt/update_rootfs_source"
-    test -d "${rootfs_dir}${update_rootfs_source}"
+    test -f "${toolbox_root_dir}${JEDI_PARTITION_TABLE_FILE}" || return 1
```
oliv3r, line 184: ditto

test/toolbox_image.sh
```diff
@@ -288,203 +110,88 @@ run_test()
...
 }
 
 usage()
 {
 cat <<-EOT
-    Usage:   "${0}" [OPTIONS] <file.img>
+    Usage:   "${0}" [OPTIONS] <toolbox image file>
```
oliv3r, line 190: :+1:

test/toolbox_image.sh
```diff
@@ -545,23 +252,11 @@ run_test test_execute_mkfs_f2fs
 run_test test_execute_resizef2fs
 run_test test_execute_mount
 run_test test_execute_rsync
-run_test test_system_update_entrypoint
+run_test test_start_update_command
+run_test test_prepare_disk_command
+run_test test_update_files_command
```
oliv3r, line 253: we may want to do this smarter, something like
```sh
test_script_command()
{
    for script in scripts/0?*.sh; do
        if [ ! -x "${toolbox_root_dir}${script}" ]; then
            echo "Warning: script '$script' not part of update toolbox, skipping'.
            continue
        fi
        chroot "${toolbox_root_dir}${script}" "-h"
    done
}
```

this does require us to define an interface for each script (every script must have the -h parameter, which is not a bad interface to agree upon).

What do you think?

oliv3r (changes requested):
as we all ready discussed, but leaving it as a comment for the far future anyway, we have to be weary that we are not building a huge test framework, when existing frameworks already exist.

Some of these changes already start going into a complete testing framework, and we don't want to have to maintain the tests And the framework

test/toolbox_image.sh
raysiudak: Yes this enforces us to implement a usage/help interface in our script which is good. 
As for you example 'I always learned' i think we should try to test only one thing in every test and make the test as descriptive as possible.

test/include/armhf_emulation.sh
raysiudak: This file provides setup and teardown functionality for arm emulation that can be used/included in a test script. I.e. it is included and functions are called from it. 
I think it is good enough like this, the file isolates the functionality, it is located in an include directory under the 'test' root folder.

test/toolbox_image.sh
raysiudak: Yes whatever is executed in the scope of the function does not get caught by the set -e of the script, therefore you need to explicitly set it in every function that you execute. 
The reason why i am doing this is because i want the tests to stop if something is wrong in the setup or teardown.

test/toolbox_image.sh
raysiudak: nope

test/toolbox_image.sh
raysiudak: Yeah for documentation purpose and consistency, although not really required because it the only statement in the function.

test/toolbox_image.sh
raysiudak: as discussed, the script is executed in a 'root' environment and the file is always readable by root. The intention of the test is to test if the file is there, other tests will fail if the content of the file is not correct.

test/include/armhf_emulation.sh
oliv3r: well for one, the overlay is not only for the ARM_EMU, i think other parts are using it as well, so that is now more generic.

Further more, it is also setting up /proc, /dev, /etc/mtab and /dev which are unrelated to the arm emulation. So it does the entire setup/teardown, not just the arm emulation specific bits. 

The comment even states this: 'because we need to write data .... temporary test files'.

it's fine the way it is in that I don't think we should split it up in smaller bits, but the naming is off :)

raysiudak: Hmmm.. ok maybe ARM_EMU naming is not nailing it. But still all the stuff done 'except the dev remounting for Docker' is done to create a virtual arm environment where the scripts can be run in. 
So therefore will rename it to...... chroot_env.sh  [dr evil pinky](https://www.google.nl/url?sa=i&rct=j&q=&esrc=s&source=images&cd=&cad=rja&uact=8&ved=2ahUKEwjzmujSzMHdAhWSaVAKHTYIAKUQjRx6BAgBEAU&url=https%3A%2F%2Ftenor.com%2Fsearch%2Fdr-evil-pinky-gifs&psig=AOvVaw0_QyQZ2im8BHYOUB6Hs8lH&ust=1537258589571672)

test/toolbox_image.sh
```diff
@@ -8,247 +8,65 @@
...
 
-    if "${is_dev_setup_mounted}"; then
-        umount "/dev" || result=1
+    if grep -q "${toolbox_root_dir}" "/proc/mounts"; then
+        umount "${toolbox_root_dir}"
     fi
 
-    if grep -q "${rootfs_dir}/proc" /proc/mounts; then
-        umount "${rootfs_dir}/proc" || result=1
+    if [ -d "${toolbox_root_dir}" ] && [ -z "${toolbox_root_dir##/tmp/*}" ]; then
```
oliv3r: i really think you should use a template to the mktemp command and check for that, mktemp does not guarantee to be using /tmp, so this may not always be true. The alternative would be to use $TMPDIR but that is empty by default, so you'd have to set it, to ensure mktemp uses it as you expect, and if so, using your own template is easier and safer.

test/include/chroot_env.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/sh
```
oliv3r, line 1: don't rename it, but test_env would have made more sense :p

test/toolbox_image.sh
raysiudak: Yes but not above set -eu

test/toolbox_image.sh
oliv3r: hmm, yeah that is true, it would have been nice to put them all under a simple template/define, hence the suggestion.
We also do this for the COMMANDS.

I'm not opposed to add 1 test per command, it _is_ good practice. but then it becomes a rather small step to also make each test a single file (or related tests together) and have one step in your gitlab-ci.yml file :)

oliv3r (commented):
awaiting rework, after which i think we can merge

test/include/chroot_env.sh
```diff
@@ -0,0 +1,109 @@
...
+    # devtmps needs to be re-mounted when running in docker
+    if ! grep -qE "/dev.*devtmpfs" "/proc/mounts"; then
+        mount -t "devtmpfs" "none" "/dev"
+        is_dev_setup_mounted=true
+    fi
+
+    # overlayfs is required because we need to write data
+    # in the ro squashfs for testing purposes,
+    # e.g. ARM_EMU_BIN mount point and temporary test files.
+    overlayfs_dir="$(mktemp -d -t "overlayfs_XXXXXX")"
```
oliv3r: to be consistent with normal mktemp usage, do "overlayfs.XXXXXXX" (useperiod rather then underscore). While we can do whatever we want, a normal mktemp is also tmp.XXXXXX so for consistency this is nicer when looking in /tmp

test/toolbox_image.sh
```diff
@@ -8,247 +8,65 @@
...
-
-    if ! grep -qE "/dev.*devtmpfs" /proc/mounts; then
-        mount -t devtmpfs none "/dev"
-        is_dev_setup_mounted=true
-    fi
-
-    mount -t devtmpfs none "${rootfs_dir}/dev"
-
-    dd if=/dev/zero of="${rootfs_dir}/${TMP_TEST_IMAGE_FILE}" bs=1 count=0 seek=128M 2> /dev/null
+    toolbox_root_dir="$(mktemp -d -t "um-update-toolbox_XXX")"
```
oliv3r: instead of _XXX use .XXX to stay conistent with normal mktemp behavior. Also, it's okay to use a bit more XXX's :) 6 when conservative, 8 being common.

oliv3r (approved):

test/toolbox_image.sh
```diff
@@ -8,247 +8,65 @@
...
-            unlink "${rootfs_dir}/${ARM_EMU_BIN}" || result=1
-        fi
+    if [ -f "${toolbox_root_dir}/${TMP_TEST_IMAGE_FILE}" ]; then
+        unlink "${toolbox_root_dir}/${TMP_TEST_IMAGE_FILE}"
     fi
 
-    if grep -q "${rootfs_dir}/dev" /proc/mounts; then
-        umount "${rootfs_dir}/dev" || result=1
-    fi
+    teardown_chroot_env "${toolbox_root_dir}"
```
oliv3r, line 46: I didn't spot this earlier, and didn't really think about this earlier, but this is a serious liability.

If, a test author, who is unaware of the consequences, puts this function in the wrong place, e.g. not at the top, you could in quite easily do:
```sh
rm -rf "${toolbox_root_dir}"
umount "${toolbox_root_dir}/proc"
```
by doing it in the wrong order, you first nuke the proc directory, and only after that try to umount it.

This can fail so horribly, in that it can bring down the server the CI runs on, as we are using privilidged mode, and mounting dev and proc from the host, rm -rf-ing those crashes the server.

Before, with the single teardown, this risk of course was also present, but it was more easy to oversee. Now, with several teardowns from each routine, this is a potential risk.

PR #42, opened 2018-09-12
Fixed verify_partition_file sanity check

raysiudak:
The check is in place for an early exit
but it did not exit


oliv3r (approved):
next time, put the error message inside the function when it fails, and just exit there.

PR #43, opened 2018-09-13
EMP-406 create prepare disk test script

raysiudak:


raysiudak:
Do not merge before other pull requests are part of master. 
This branch should remove the first commit 3f91991 and rebase on master first

.gitlab-ci.yml
```diff
@@ -162,6 +163,7 @@ build:master_docker_latest:
     - build:master_docker_latest
   script:
     - ./test/toolbox_image.sh .build_*/um-update*.xz.img
+    - ./test/prepare_disk.sh .build_*/um-update*.xz.img
```
oliv3r: for our tests, i'd recommend putting each in its own job. feel free to template it so it only reads:
```yaml
test:toolboxtest_1/2:
<<: *job_template
script: ./test/toolbox_image.sh ...
```
or something. a) it allows for the tests to run in parallel; b) you more easily see which tests fails in the overview, rather then 'all tests failed'

test/include/armhf_emulation.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/sh
```
oliv3r: i think this was part of the 417 review so ignoring it here

test/prepare_disk.sh
```diff
@@ -0,0 +1,431 @@
+#!/bin/sh
```
oliv3r, line 1: you reccon it a good idea to name the script scripts/30_prepare_disk here to improve searchability etc?

test/prepare_disk.sh
```diff
@@ -0,0 +1,431 @@
...
+
+toolbox_image=""
+toolbox_root_dir=""
+work_dir=""
+
+exit_on_failure=false
+result=0
+
+# shellcheck source=test/include/armhf_emulation.sh
+. "test/include/armhf_emulation.sh"
```
oliv3r: put the include at the top of the file, ideally before the set -eu bit

test/prepare_disk.sh
```diff
@@ -0,0 +1,431 @@
...
+
+
+execute_prepare_disk()
```
oliv3r: i assume everything below is just a copy/paste from the other script, so ignoring that

test/prepare_disk.sh
```diff
@@ -0,0 +1,431 @@
...
+
+    # Remove the name from the source partition file, as the names are not stored in the partition table.
+    sed -i "s/, name=.*//" "${PARTITION_TABLE_FILE}"
+
+    diff -b "${PARTITION_TABLE_FILE}" "${PARTITION_TABLE_FILE}.verify" || return 1
+}
+
+test_disk_integrity()
+{
+    set -e
```
oliv3r: hmm, this looks like a change, and why is that needed on a per test basis? Either the set -eu (what happened to the u) is applied whenever the file is sourced, or from the sourcing file, i'd be supprised that you have to set it each call to the function. (Unless you are doing if test_function; then) then, i understand why you are doing this, but then we should really re-evaluate our testing framework.

oliv3r (changes requested):

.gitlab-ci.yml
raysiudak: Yes we could do that, but.... this will result in having to add the tests in the yaml file for every build job type, i.e. 

test:dev_docker_candidate
test:dev_docker_latest
test:master_docker_latest

So both solutions are not ideal, lets park this for now in EMP-374 and pick this up when we setup all the repositories properly.

test/prepare_disk.sh
raysiudak: No i don't think so, i think adding the number for sorting should only be added if this is intended and not for the reason of script names having the exactly tje same filename.

test/prepare_disk.sh
raysiudak: I moved this to the top, but it should be under the 'set' statement because you want the including script to control this behaviour and the include script introduces variables that potentially be unset.

test/prepare_disk.sh
raysiudak: Yes this is wrong. The set statement is now only in the top of the file before everything else, all functions expected to exit on error are not executed in an 'if' statement. 
Tests however are executed within an 'if' statement and should return something

test/prepare_disk.sh
raysiudak: Yes except for some variable naming and the location of the temporary files.

test/include/chroot_env.sh
```diff
@@ -27,7 +27,7 @@ setup_chroot_env()
     # overlayfs is required because we need to write data
     # in the ro squashfs for testing purposes,
     # e.g. ARM_EMU_BIN mount point and temporary test files.
-    overlayfs_dir="$(mktemp -d -t "overlayfs.XXXXXXXXXX")"
+    overlayfs_dir="$(mktemp -d -t "overlayfs.XXXXXX")"
```
oliv3r: I think this was an unintended change whilst resolving.

.gitlab-ci.yml
```diff
@@ -162,7 +162,8 @@ build:master_docker_latest:
     - build:dev_docker_latest
     - build:master_docker_latest
   script:
-    - ./test/toolbox_image.sh ".build_"*"/um-update"*".xz.img"
+    - ./test/toolbox_image.sh .build_*/um-update*.xz.img
+    - ./test/prepare_disk.sh .build_*/um-update*.xz.img
```
oliv3r: I think this was an inintended change from merging, also also quote the new addition :)

test/include/chroot_env.sh
```diff
@@ -76,13 +76,13 @@ teardown_chroot_env()
...
 
-    mounts="${overlayfs_dir}/rom ${overlayfs_dir}"
+    mounts="${overlayfs_dir}/rom ${overlayfs_dir} ${target_root_dir}"
     for mount in ${mounts}; do
         if grep -q "${mount}" "/proc/mounts"; then
             umount "${mount}"
         fi
-        if [ -d "${mount}" ] && [ -z "${mount##*overlayfs_*}" ]; then
-            rm -r "${mount:?}"
+        if [ -d "${mount}" ] && [ -z "${mount##*overlayfs*}" ]; then
```
oliv3r: p.s. you could also put your template/mask in a variable and reference it in the needed spots :p (just a tip, :tipping_hand_man:)

test/prepare_disk.sh
```diff
@@ -0,0 +1,431 @@
...
+    ?)
+        echo "Invalid option: -${OPTARG}."
+        exit 1
+        ;;
+    esac
+done
+shift "$((OPTIND - 1))"
+
+if [ "${#}" -ne 1 ]; then
+    echo "Missing argument <file.img>."
```
oliv3r: Hmm, only spotted this by accident here, but did you not rename this before?

oliv3r (changes requested):

test/include/chroot_env.sh
raysiudak: This is intended, six random positions are enough

.gitlab-ci.yml
raysiudak: This is not intended.

test/prepare_disk.sh
raysiudak: I changed it in the usage, not in this place

test/include/chroot_env.sh
oliv3r: i'll mark it as resolved, sine this will be merged via EMP-407

.gitlab-ci.yml
oliv3r: as this was also fixed in EMP-407, marking as resolved.

test/include/chroot_env.sh
oliv3r: ditto, per EMP-407

test/prepare_disk.sh
oliv3r: Not sure this was changed here, EMP-407 did change it elsewhere I think

test/prepare_disk.sh
```diff
@@ -0,0 +1,254 @@
...
+    mkfs.f2fs -q "${TARGET_STORAGE_DEVICE}p3"
+
+    test_disk_integrity
+
+    echo "Successfully created dummy storage device: '${TARGET_STORAGE_DEVICE}'."
+}
+
+setup()
+{
+    toolbox_root_dir="$(mktemp -d -t "um-update-toolbox.XXXXXXX")"
```
oliv3r: why did you not use the name template that you used with the update_files script? would have been nice and consistent :p

.gitlab-ci.yml
```diff
@@ -163,8 +163,8 @@ build:master_docker_latest:
     - build:master_docker_latest
   script:
     - ./test/toolbox_image.sh ".build_"*"/um-update"*".xz.img"
-    - ./test/update_files.sh ".build_"*"/um-update"*".xz.img"
     - ./test/prepare_disk.sh ".build_"*"/um-update"*".xz.img"
+    - ./test/update_files.sh ".build_"*"/um-update"*".xz.img"
```
oliv3r: since this is part of your first commit, you could have just fixed it there.

test/include/chroot_env.sh
```diff
@@ -29,7 +29,7 @@ setup_chroot_env()
     # overlayfs is required because we need to write data
     # in the ro squashfs for testing purposes,
     # e.g. ARM_EMU_BIN mount point and temporary test files.
-    overlayfs_dir="$(mktemp -d -t "${NAME_TEMPLATE_OVERLAYFS}.XXXXXXXX")"
+    overlayfs_dir="$(mktemp -d -t "${NAME_TEMPLATE_OVERLAYFS}.XXXXXX")"
```
oliv3r, line 32: this is my bad, i fixed it in 407, but messed up on that eventually, sorry.

test/prepare_disk.sh
```diff
@@ -126,13 +126,13 @@ teardown()
 
     cd "${CWD}"
 
-    if [ -d "${work_dir}" ] && [ -z "${dir##*prepare_disk_work_dir_*}" ]; then
+    if [ -d "${work_dir}" ] && [ -z "${work_dir##*prepare_disk_work_dir*}" ]; then
```
oliv3r: this fix isn't part of a merge conflict or this commit!

test/prepare_disk.sh
```diff
@@ -379,7 +379,7 @@ done
 shift "$((OPTIND - 1))"
 
 if [ "${#}" -ne 1 ]; then
-    echo "Missing argument <file.img>."
+    echo "Missing argument <toolbox image file>."
```
oliv3r: sneaked this change unrelated to this changeset in here :) and completly unessasery as you added this in the first commit of this changeset.

oliv3r (commented):
The commit hygiene 72da14d548e402a9b ("Removed script name from test function, since this test scripts is 
intended to run tests for only that script.") is in need of some improvement.

raysiudak:
I am confused about what the state is of this pull request, what is still requried.

oliv3r (approved):

PR #44, opened 2018-09-13
EMP-407 create update files script test

raysiudak:


scripts/50_update_files.sh
```diff
@@ -82,15 +82,32 @@ if [ -z "${UPDATE_ROOTFS_SOURCE}" ] || [ -z "${TARGET_STORAGE_DEVICE}" ]; then
 fi
 
 if [ ! -d "${UPDATE_ROOTFS_SOURCE}" ]; then
-    echo "Update failed: ${UPDATE_ROOTFS_SOURCE} does not exist."
+    echo "Update failed: '${UPDATE_ROOTFS_SOURCE}' does not exist."
+    exit 1
+fi
+
+if ! cat "${UPDATE_ROOTFS_SOURCE}/etc/debian_version" > /dev/null 2>&1; then
```
oliv3r: i thought 2>&1 was not allowed, i'd keep the 2> case though, IF it fails, you atleast know WHY.

Btw, does it matter if its a debian version? as long as its an ultimaker version, its fine. If we switch to buildroot or whatnot, this would be the only test holding us back. I don't think it's relevant.

test/include/armhf_emulation.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/sh
```
oliv3r: the other reviews cover this file i'm sure

test/update_files.sh
```diff
@@ -0,0 +1,295 @@
+#!/bin/sh
```
oliv3r, line 1: i assume this is just a copy paste like with the other cases. so ignoring this file

oliv3r (changes requested):

scripts/50_update_files.sh
raysiudak: If we change the distribution this test will fail and that tells us that we need to implement a check again for a different os and forces us to at least have a look at the sanity checks.

oliv3r: i do not disagree entirely, but this now becomes an interface, that we at the least need to document somewhere.

oliv3r (approved):

PR #46, opened 2018-09-18
EMP-307 Update README documentation

raysiudak:
Add information about how the system configuration is done,
what scripts are there, what configuration files, how
they relate and how they are executed.

Contributes to EMP-272 and EMP-307

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,30 +1,43 @@
 # Introduction
-The 'um-update_toolbox_armhf' is a 'squashfs' minimal root filesystem that can be mounted in early boot stages, with the purpose of providing tools to be able to perform complex setup and update operations. 
+The *um-update_toolbox* is a *squashfs* minimal root filesystem that can be mounted in early boot stages, with the purpose of providing tools to be able to perform complex setup and update operations. 
 
-This toolbox image is based on Linux Alpine, a security-oriented, lightweight Linux distribution based on musl libc and busybox. 
+This toolbox image is based on *Linux Alpine*, a security-oriented, lightweight Linux distribution based on *Musl Libc* and *Busybox*. 
 
-On top of the standard distro we install the required tooling, such as; e2fsprogs-extra f2fs-tools rsync, resulting in an image with a compressed size of less then 2.6M at time of writing. 
+On top of the standard *distro* we install the required tooling, such as; *e2fsprogs-extra f2fs-tools rsync*, resulting in an image with a compressed size of less then 2.6M at time of writing. 
 
 ## Repository structure
 #### Repo root directory:
 - License and readme files
-- Docker files used by CI for generating an isolated build environment see...
+- Docker files used by *CI* for generating an isolated build environment.
 - Scripts for building the software
-- Configuration files for Gitlab CI
+- Configuration files for *Gitlab CI*
+
+#### Config directory:
+This directory contains configuration files that are used by the
+system configuration scripts in the 'scripts' directory. The config files
+are installed in the '/etc/system_update'.
 
 #### Scripts directory:
-
-Update or setup scripts, where the main entrypoint is startup.sh. These 
-scripts will be placed in the toolbox '/sbin' folder  
+Contains update and setup scripts. The main entrypoint is 'start_update.sh' which is installed
+ in the system binary directory '/sbin', note that this is a hardcoded dependency for the 
+ *um-kernel* repository since the *initramfs* init script looks for this script. The other
+ configuration scripts are installed in the '/etc/system_update.d'.    
 
 #### Tests directory:
 - A test to check the build environment
 - A test to validate the generated toolbox functionality
+- Unit tests to validate the system configuration scripts. 
 
 ## How to build
-This repository is setup to use an existing docker image from our docker registry at gitlab. The image contains all the tools to be able to build the software. When docker is not installed the build script will try to build on the host machine and will probably fail because the required tools are not installed. When it is necessary to run the build without docker, execute the tests/buildenv.sh script and see if the environment is missing requirements. 
+This repository is setup to use an existing *Docker* image from this repositories Docker registry. 
+The image contains all the tools to be able to build the software. When Docker is not installed 
+the build script will try to build on the host machine and is likely to fail because the required 
+tools are not installed. When it is necessary to run the build without Docker, execute the 
+'tests/buildenv.sh' script and see if the environment is missing requirements. 
 
-By default the build script runs an environment check, builds the image and then validates it by running tests. The first and the latter can be disabled. Run the help of the build_for_ultimaker.sh script for usage information:  
+By default the build script runs an environment check, builds the image and then validates 
+it by running tests. The first and the latter can be disabled. Run the help of the 
+'build_for_ultimaker.sh' script for usage information:  
 ```sh
 > ./build_for_ultimaker.sh -h
     Usage: ./build_for_ultimaker.sh [OPTIONS]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,23 +8,35 @@ On top of the standard distro we install the required tooling, such as; e2fsprog
 ## Repository structure
 #### Repo root directory:
 - License and readme files
-- Docker files used by CI for generating an isolated build environment see...
+- Docker files used by CI for generating an isolated build environment see: [Software CI/CD](https://confluence.ultimaker.com:8443/x/ybC9)
```
oliv3r: while it's super useful, I'm not sure how wise it is to put 'private, hidden' documentation in confluence via a link as such. I'd remove the 'see: .....' bit.

README.md
```diff
@@ -8,23 +8,35 @@ On top of the standard distro we install the required tooling, such as; e2fsprog
...
 - Scripts for building the software
 - Configuration files for Gitlab CI
 
-#### Scripts directory:
+#### Config directory:
+This directory contains configuration files that are used by the
+system configuration scripts in the 'scripts' directory. The scripts
+are installed in the '/etc/system_update' AKA the 'system update directory'.
```
oliv3r: don't change it here, i have it in EMP-411 for now, but the script should not be in /etc, /etc holds config files, not scripts, the scripts are in /usr/libexec/jedi_system_update/.

But I'm not sure what you mean with the 'AKA" bit. I don't think we have a system update directory or something known was.

README.md
```diff
@@ -8,23 +8,35 @@ On top of the standard distro we install the required tooling, such as; e2fsprog
...
 
-Update or setup scripts, where the main entrypoint is startup.sh. These 
-scripts will be placed in the toolbox '/sbin' folder  
+#### Scripts directory:
+Contains update or setup scripts. The main entrypoint is start_update.sh which is installed
```
oliv3r: update and setup

not update or setup, as it contains both, not either :)

"start_update.sh" maybe make it bold or italic to let it stand out? Directories are single quoted, maybe only single quotes is enough?

README.md
```diff
@@ -8,23 +8,35 @@ On top of the standard distro we install the required tooling, such as; e2fsprog
...
+ in the system binary directory /sbin, note that this is a hardcoded dependency for the
```
oliv3r: single qute /sbin

README.md
```diff
@@ -8,23 +8,35 @@ On top of the standard distro we install the required tooling, such as; e2fsprog
...
+This directory contains configuration files that are used by the
+system configuration scripts in the 'scripts' directory. The scripts
+are installed in the '/etc/system_update' AKA the 'system update directory'.
 
-Update or setup scripts, where the main entrypoint is startup.sh. These 
-scripts will be placed in the toolbox '/sbin' folder  
+#### Scripts directory:
+Contains update or setup scripts. The main entrypoint is start_update.sh which is installed
+ in the system binary directory /sbin, note that this is a hardcoded dependency for the 
+ um-kernel repository since the initrd init script looks for this script.
```
oliv3r: note to self: we need to document this as an interface

README.md
```diff
@@ -8,23 +8,35 @@ On top of the standard distro we install the required tooling, such as; e2fsprog
...
-scripts will be placed in the toolbox '/sbin' folder  
+#### Scripts directory:
+Contains update or setup scripts. The main entrypoint is start_update.sh which is installed
+ in the system binary directory /sbin, note that this is a hardcoded dependency for the 
+ um-kernel repository since the initrd init script looks for this script.    
 
 #### Tests directory:
 - A test to check the build environment
 - A test to validate the generated toolbox functionality
+- Unit tests to validate the system configuration scripts.
```
oliv3r, line 29: i'd just call it system scripts. (As currently is remotely implies the scripts that read the configuration files or something).

README.md
```diff
@@ -8,23 +8,35 @@ On top of the standard distro we install the required tooling, such as; e2fsprog
...
+ um-kernel repository since the initrd init script looks for this script.    
 
 #### Tests directory:
 - A test to check the build environment
 - A test to validate the generated toolbox functionality
+- Unit tests to validate the system configuration scripts. 
 
 ## How to build
-This repository is setup to use an existing docker image from our docker registry at gitlab. The image contains all the tools to be able to build the software. When docker is not installed the build script will try to build on the host machine and will probably fail because the required tools are not installed. When it is necessary to run the build without docker, execute the tests/buildenv.sh script and see if the environment is missing requirements. 
+This repository is setup to use an existing docker image from our docker registry at gitlab.
```
oliv3r: 'from this repositories docker registry'

otherwise, this implies we have some global registry setup at gitlab, which we do not.

README.md
```diff
@@ -8,23 +8,35 @@ On top of the standard distro we install the required tooling, such as; e2fsprog
...
 #### Tests directory:
 - A test to check the build environment
 - A test to validate the generated toolbox functionality
+- Unit tests to validate the system configuration scripts. 
 
 ## How to build
-This repository is setup to use an existing docker image from our docker registry at gitlab. The image contains all the tools to be able to build the software. When docker is not installed the build script will try to build on the host machine and will probably fail because the required tools are not installed. When it is necessary to run the build without docker, execute the tests/buildenv.sh script and see if the environment is missing requirements. 
+This repository is setup to use an existing docker image from our docker registry at gitlab. 
+The image contains all the tools to be able to build the software. When docker is not installed 
+the build script will try to build on the host machine and will probably fail because the required
```
oliv3r: machine and may fail because

or

machine and is likely to fail

README.md
```diff
@@ -34,11 +46,40 @@ By default the build script runs an environment check, builds the image and then
 ```
 
 ## Adding update or setup routines to the toolbox
-During the initramfs routine (executed between 2st stage U-boot and 3th stage Linux debian load) this toolbox image mounted and the executable entryscript '/sbin/startup.sh' will be executed. This script shall be used to execute any required routine. Very important here is to make sure all required resources are nicely cleaned up and errors are properly handled, not to bring the system in an unrecoverable state. 
+During the initramfs routine (executed between 2st stage U-boot and 3th stage Linux debian load)
```
oliv3r: During the initramfs stage may actually be better?
as we do not have an initramfs routine, but rather, run the init routine from within the initramfs.

In any case, i'd even drop the commented bit, (which btw is 2nd stage and 3rd rather). If we say it runs from the kernels initramfs stage, we should have said enough.

What we can do of course, add a nice graph on confluence which explains all these steps with links to all involved repositories/files

README.md
```diff
@@ -34,11 +46,40 @@ By default the build script runs an environment check, builds the image and then
 ```
 
 ## Adding update or setup routines to the toolbox
-During the initramfs routine (executed between 2st stage U-boot and 3th stage Linux debian load) this toolbox image mounted and the executable entryscript '/sbin/startup.sh' will be executed. This script shall be used to execute any required routine. Very important here is to make sure all required resources are nicely cleaned up and errors are properly handled, not to bring the system in an unrecoverable state. 
+During the initramfs routine (executed between 2st stage U-boot and 3th stage Linux debian load) 
+this toolbox image mounted and the executable entryscript '/sbin/start_update.sh' will be executed.
```
oliv3r: image is mounted

will be executed if available.

README.md
```diff
@@ -34,11 +46,40 @@ By default the build script runs an environment check, builds the image and then
...
+ 
+The 'system configuration' scripts are installed in /etc/system_update.d folder and are prefixed
```
oliv3r: The system update scripts are ...
no need to quote this

folder -> directory
that's why you use the .d parameter :p as that stands for directory :)

README.md
```diff
@@ -34,11 +46,40 @@ By default the build script runs an environment check, builds the image and then
 ```
 
 ## Adding update or setup routines to the toolbox
-During the initramfs routine (executed between 2st stage U-boot and 3th stage Linux debian load) this toolbox image mounted and the executable entryscript '/sbin/startup.sh' will be executed. This script shall be used to execute any required routine. Very important here is to make sure all required resources are nicely cleaned up and errors are properly handled, not to bring the system in an unrecoverable state. 
+During the initramfs routine (executed between 2st stage U-boot and 3th stage Linux debian load) 
+this toolbox image mounted and the executable entryscript '/sbin/start_update.sh' will be executed.
+ 
+The 'system configuration' scripts are installed in /etc/system_update.d folder and are prefixed 
+with a two digit number so the 'start_update' script can sort them.
```
oliv3r: 'start_update.sh'
and

script can execute them in order.

The script won't sort anything :p

README.md
```diff
@@ -34,11 +46,40 @@ By default the build script runs an environment check, builds the image and then
...
+ 
+A chrooted environment will be prepared with the 'sysroot' files of the 'um-update_toolbox' files,
```
oliv3r: What do you mean with 'sysroot'? Commonly sysroot refers to a cross-build system root.
The 'um-update_toolbox' will serve as a chroot environment for the update ...
is fine I think.

README.md
```diff
@@ -34,11 +46,40 @@ By default the build script runs an environment check, builds the image and then
...
+all scripts use te same runtime environment in which the required variables are defined.
```
oliv3r: I think we forget to mention WHY we need this/want this, which should probably be somewhere in the intro section.

README.md
```diff
@@ -34,11 +46,40 @@ By default the build script runs an environment check, builds the image and then
...
+
+* TARGET_STORAGE_DEVICE    - The device to perform the system configuration on
+* PARTITION_TABLE_FILE     - The partition table file used in the prepare_disk script,
+                             default is 'jedi_emmc_sfdisk.table', implicitly the 
+                             script will look for a counterpart checksum file:
+                             '<filename>.sha512'
+* UPDATE_EXCLUDE_LIST_FILE - A file containing all files and directories to be excluded
+                             used by the 'update_files script', default is 'jedi_update_exclude_list.txt'
+* UPDATE_ROOTFS_SOURCE     - This is the directory within the 'toolbox sysroot' that has the update files,
+                             this is used in the update_files script as the source directory for the
```
oliv3r: update files. This is used

README.md
```diff
@@ -34,11 +46,40 @@ By default the build script runs an environment check, builds the image and then
...
+#### Important environment variables 
+
+* TARGET_STORAGE_DEVICE    - The device to perform the system configuration on
+* PARTITION_TABLE_FILE     - The partition table file used in the prepare_disk script,
+                             default is 'jedi_emmc_sfdisk.table', implicitly the 
+                             script will look for a counterpart checksum file:
+                             '<filename>.sha512'
+* UPDATE_EXCLUDE_LIST_FILE - A file containing all files and directories to be excluded
+                             used by the 'update_files script', default is 'jedi_update_exclude_list.txt'
+* UPDATE_ROOTFS_SOURCE     - This is the directory within the 'toolbox sysroot' that has the update files,
```
oliv3r: This is the directory within the chroot environment (or toolbox environment) that contains the update files

README.md
```diff
@@ -34,11 +46,40 @@ By default the build script runs an environment check, builds the image and then
...
+                             this is used in the update_files script as the source directory for the
+                             update files. 
+* SYSTEM_UPDATE_DIR        - The system update directory is a directory within the 'toolbox sysroot' 
+                             that contains the system configuration scripts. The configuration files 
+                             can be found in '/etc/system_update', i.e. the partition table and exclude
+                             list file. The system configuration scripts can be found in the same 
+                             directory post-fixed with '.d' (/etc/system_update.d).
```
oliv3r: It looks like the variables should be somewhere at the top, and then referenced int he rest of the document.

e.g. you now reference the /etc/system_update directory 2 or 3 times, whereas instead, it would have been nice to refer to $SYSTEM_UPDATE_DIR instead

oliv3r (changes requested):
Overal, looks good, just some small things

README.md
raysiudak: Actually it does sort (well globbing sorts) before executing.

README.md
raysiudak: Hmm.. yeah, could be, but then the whole build up would change. I will keep it like it is.

README.md
oliv3r: lol that, is true, it does sort them while executing, but not sort them in place/store them sorted.

oliv3r (approved):

PR #49, opened 2018-09-26
Minor fixes/improvements to the update_files script

oliv3r:
- scripts/update_files: Use safe template cleanup pattern
- scripts/update_files: Avoid the use of and/or
- scripts/update_files: Print usage in case of missing parameters

### DIFF
--- a/scripts/50_update_files.sh
+++ b/scripts/50_update_files.sh
@@ -106,9 +106,13 @@ while getopts ":d:hs:" options; do
 done
 shift "$((OPTIND - 1))"
 
-if [ -z "${UPDATE_ROOTFS_SOURCE}" ] || [ -z "${TARGET_STORAGE_DEVICE}" ]; then
-    echo "Missing arguments <UPDATE_ROOTFS_SOURCE> and/or <TARGET_STORAGE_DEVICE>."
-    usage
+if [ -z "${UPDATE_ROOTFS_SOURCE}" ]; then
+    echo "Missing arguments <UPDATE_ROOTFS_SOURCE>."
+    exit 1
+fi
+
+if [ -z "${TARGET_STORAGE_DEVICE}" ]; then
+    echo "Missing arguments <TARGET_STORAGE_DEVICE>."
     exit 1
 fi
 
--- a/scripts/50_update_files.sh
+++ b/scripts/50_update_files.sh
@@ -7,6 +7,8 @@
 
 set -eu
 
+trap cleanup EXIT
+
 # common directory variables
 SYSCONFDIR="${SYSCONFDIR:-/etc}"
 
@@ -17,7 +19,8 @@ TARGET_STORAGE_DEVICE="${TARGET_STORAGE_DEVICE:-}"
 UPDATE_ROOTFS_SOURCE="${UPDATE_ROOTFS_SOURCE:-}"
 # end system_update wide configuration settings
 
-UPDATE_TARGET="/tmp/target_root"
+NAME_TEMPLATE_UPDATE_TARGET="um-target_root"
+UPDATE_TARGET="$(mktemp -d -t "${NAME_TEMPLATE_UPDATE_TARGET}.XXXXXX")"
 
 usage()
 {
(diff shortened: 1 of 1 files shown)


scripts/50_update_files.sh
```diff
@@ -32,10 +35,39 @@ usage()
...
 
+cleanup()
+{
+    # On slow media, umount and/or rmdir can fail with 'resource busy' errors.
+    # To do our best with cleanup, attempt this a few times before giving up.
+    failed=0
+    while test -d "${UPDATE_TARGET}"; do
+        if grep -q "${UPDATE_TARGET}" "/proc/mounts"; then
+            if ! umount "${UPDATE_TARGET}"; then
+                failed=$((failed + 1))
```
raysiudak: quotes

scripts/50_update_files.sh
```diff
@@ -32,10 +35,39 @@ usage()
...
+            fi
+        fi
+
+        if [ -d "${UPDATE_TARGET}" ] && \
+           [ -z "${UPDATE_TARGET##*${NAME_TEMPLATE_UPDATE_TARGET}*}" ]; then
+            if ! rmdir "${UPDATE_TARGET}"; then
+                failed=$((failed + 1))
```
raysiudak: quotes

scripts/50_update_files.sh
raysiudak: Could it happen that the script never returns?
In that case it could be safer to let a reboot clean up (because of temp) the mess if the umount and/or delete fails.

raysiudak (changes requested):

scripts/50_update_files.sh
oliv3r: are you sure? I omitted them because i remember shell check complaining about that, as it's an integer, not a string. But going to shellcheck.net it doesn't care either way. So i'll add them for consistency. My memory is getting bad!

scripts/50_update_files.sh
oliv3r: it is possible for this to fail, if the device stays 'busy' for more then 15 to 30 seconds. The update script then exists with error 1.

so the question then is, do we want to increase the '-ge 30' to allow for more time on failure (which should recover eventually to be fair) (i increase it to 300 for max 10 minutes, in case the flash needs to write a lot) or what else?

The rsync + sync job should try to wait as long as possible, but if the MMC module crashes and just waits for ever. What is the safest step to continue with? Reboot probably not, as a reboot will leave the system in an unknown state. That's for start_update.sh to handle determine in case of a failed function ...

p.s. umount does the 'sync' part, but the flash controller may lie a little bit (but not for huge amounts of writes) .. e.g.
user: 'write 100mb' 
flash: 'ok writing ...'
flash: 'done!' (but i'm really not done, just that my buffers are full, so I could be done)
user: 'write 200mb'
flash: 'wow wait up, i'm not done yet' (still writing my buffers to flash, this flash wear leveling stuff is taking ages)

raysiudak: The question is should the script fail if temporary stuff can not be cleaned up properly, although we gave it a good try, because of limitations in the os?

oliv3r: you missunderstand, this is not temporary stuff not being up properly, this is umount. And umount is failing because the device it is trying to umount is still busy.
Busy being very likely that its int he background still writing to the actual flash.

So reboot while the controller is still doing that, loose changes, have an undefined state on your filesystem, that you don't even know (corrupt files etc).

The OS is just trying to be helpful here, in letting you know something is not right yet.

remember, that the 'temporary' stuff is just the mount point, which points to an actual root disk.

raysiudak: Ah yes you are right, but shouldn't you only try to delete the mount point then when the actual umount succeeds?

oliv3r: well since we are using rmdir, it will only work if it is not mounted :)

alternatively, we could add a sleep 1; continue after failed++, so that it does not try rmdir if you want. I wanted to avoid the duplication of the sleep 1 is all

raysiudak (approved):

PR #50, opened 2018-09-27
WIP: EMP-274 phase5 save and restore

oliv3r:


test/toolbox_image.sh
```diff
@@ -185,7 +184,7 @@ test_update_files_command()
 
 test_jedi_exclude_list_exists()
 {
-    test -f "${toolbox_root_dir}/${JEDI_EXCLUDE_LIST_FILE}" || return 1
+    test -d "${toolbox_root_dir}/${SYSTEM_UPDATE_CONF_DIR}" || return 1
```
raysiudak, line 187: This test just doesn't make that much sense now since it is not an input anymore and we do validate that the filter works correctly. But we never check that the filter 'filters' what we explicitly wan't to filter.

scripts/50_update_files.sh
```diff
@@ -142,8 +150,14 @@ if [ ! -b "${TARGET_STORAGE_DEVICE}" ]; then
     exit 1
 fi
 
-if [ ! -f "${SYSTEM_UPDATE_CONF_DIR}/${UPDATE_EXCLUDE_LIST_FILE}" ]; then
-    echo "Update failed: file '${SYSTEM_UPDATE_CONF_DIR}/${UPDATE_EXCLUDE_LIST_FILE}' not found."
+if [ ! -d "${SYSTEM_UPDATE_CONF_DIR}" ]; then
```
raysiudak, line 153: I do not understand why there are multiple locations here and above to loop for filter files?

scripts/20_save_data.sh
```diff
@@ -0,0 +1,164 @@
...
+fi
+
+if [ ! -d "${UPDATE_SAVE_DIR}" ]; then
+    echo "Missing directory to save files to, cannot continue."
+    usage
+    exit 1
+fi
+
+if [ ! -f "${KEEP_LIST}" ]; then
+    echo "Nothing to save according to '${SYSTEM_UPDATE_CONF_DIR}'."
```
raysiudak, line 157: This message is a little confusing, the configuration dir does not determine anything. I suggest something in the line of: 
echo "No backup required, no '.keep' files found in '${SYSTEM_UPDATE_CONF_DIR}'."

scripts/20_save_data.sh
```diff
@@ -0,0 +1,164 @@
...
+            continue
+        fi
+
+        # Tar likes its exclude list to be relative, so replace the initial
+        # '/' on a line with './'.
+        sed s,^/,./, "${keep}" >> "${KEEP_LIST}"
+    done
+
+    for partition in "${TARGET_STORAGE_DEVICE}"*; do
+        if [ ! -b "${partition}" ] || \
```
raysiudak, line 83: Remove || and be verbose

raysiudak: You hate 2> /dev/null

scripts/20_save_data.sh
```diff
@@ -0,0 +1,164 @@
...
+        update_target="${TARGET_MOUNT}"
+
+        # When using overlay's, the actual data is stored in the 'upper'
+        # sub-directory, as such, we use that as our 'root' directory.
+        if [ -d "${TARGET_MOUNT}/upper" ]; then
+            update_target="${TARGET_MOUNT}/upper"
+        fi
+
+        update_archive="${UPDATE_SAVE_DIR}/$(basename "${partition}").tar.xz"
+        tar --ignore-failed-read -cJf "${update_archive}" -C "${update_target}" --files-from "${KEEP_LIST}" 2> /dev/null
```
raysiudak, line 97: This is probably the problematic 2> /dev/null :P

scripts/20_save_data.sh
```diff
@@ -0,0 +1,164 @@
...
+
+KEEP_LIST="$(mktemp)"
+
+NAME_TEMPLATE_TARGET_MOUNT="um-save_mount"
+TARGET_MOUNT="$(mktemp -d -t "${NAME_TEMPLATE_TARGET_MOUNT}.XXXXXX")"
+
+usage()
+{
+    echo "Usage: ${0} [OPTIONS]"
+    echo "Save the data files from the TARGET_STORAGE_DEVICE."
```
raysiudak, line 27: Well you could say all files are data, but that means that everything is saved from the TARGET_STORAGE_DEVICE. Suggest: Save (Why not backup) files listed in the .keep files found in the <SYSTEM_CONFIG_DIR>

scripts/20_save_data.sh
```diff
@@ -0,0 +1,164 @@
+#!/bin/sh
```
raysiudak, line 1: Why did you choose to name the file save_data, i think backup much better describes the intended behaviour of this file

raysiudak (changes requested):

scripts/20_save_data.sh
oliv3r: i did have it backup first, but I explicitly decided NOT to name it backup, as we are not actually backing up anything. After a reboot the 'backup' is gone. So I wanted to make sure it did not give a sense of having a backup for later usage.

But i"m very much open to better naming then backup.
(I borrowed the name save_data from openwrt where they also save the cfg data between updates)

scripts/20_save_data.sh
oliv3r: you are right, I don't want to use 2> /dev/null anymore.
Here I thought it may not be bad, as we may get quite a few errors from it (any partition that cannot be mounted will produce an error, even if that's okay).

As for the removing the or, I'm not sure I understand why

scripts/20_save_data.sh
oliv3r: yes! it is gone since.

scripts/50_update_files.sh
```diff
@@ -75,8 +74,17 @@ perform_update()
...
-    if ! rsync --exclude-from "${SYSTEM_UPDATE_CONF_DIR}/${UPDATE_EXCLUDE_LIST_FILE}" -c -a --delete -x \
-        "${UPDATE_ROOTFS_SOURCE}/" "${UPDATE_TARGET}/"; then
+    # Note that with rsync --delete, we want to ensure both the '.keep' as
+    # well as the '.discard' file are ignored.
+    for exclude in "${SYSTEM_UPDATE_CONF_DIR}/"*".keep" \
+                   "${SYSTEM_UPDATE_CONF_DIR}/"*".discard" \
+                   "${UPDATE_ROOTFS_SOURCE}/${SYSTEM_UPDATE_CONF_DIR}"*".keep" \
+                   "${UPDATE_ROOTFS_SOURCE}/${SYSTEM_UPDATE_CONF_DIR}"*".discard"; do
+        exclude_list="${exclude_list:-} --exclude-from ${exclude}"
+    done
```
oliv3r, line 84: We are checking for both locations, in /toolbox/etc/jedi-system/config and in the mounted root file system. This to ensure that if an application adds keep files (which they should) these are also honored. Actually, the griffin, cluster etc stuff should install their own keep files. What we have now is a shortcut! by putting into the toolbox.

The discard files are only for rsync and only for internal usage. So maybe the "${UPDATE_ROOTFS_SOURCE}/${SYSTEM_UPDATE_CONF_DIR}"*".discard" can go (but i like the symetry). Once we switch to a squashfs root image, we can drop the discard files altogether.

scripts/50_update_files.sh
oliv3r: you are right, this check should go, as the filter loop accumulates the needed files.

oliv3r:
Closed via ultimaker/jedi-system-update#14

